PR #3, opened 2019-04-02
replaced mkdir with makedirs to create all the intermediate directories if they don't exist

SebGue:


BonneelP:

PR #5, opened 2019-04-10
Added some class descriptions

SebGue:
Hello Pierre,

I added some class descriptions to my best knowledge. 
I also removed input parameter of build_geometry(Rbo) in magnet.plot() since it seems to be unused and causes an error if one try to plot magnets alone.

Best regards, Sebastian


BonneelP:
Thank you for your help !

Before committing, can you use black on the project ? In your last commit there are 129 files modified, most of them are due to the fact that the class generator doesn't generate PEP8 class files. It hides the interesting part of your commits. I will open an issue on that.

BonneelP:
Hello Sebastian, 

It seems that there was an error during your merge (before commit "Added some class descriptions") . Most of your modification are not in the pull request. It seems that there is only the modification related to the class descriptions. Can you check what happened ? 

Best regards,
Pierre

SebGue:
Hello Pierre,

I checked my commit and it seems to be okay (ie. Methods/Machine/Magnet/plot.py is also commited). It is the last file in the diff comparision.
I'm sorry for me beeing to lazy to only include the changed class files.

Best regards,
Sebastian

BonneelP:
No problem, I though that you erased the FEMM related commit by mistake. My bad. Take as long as you need ;)

SebGue:
Now I know what missing content you meant. I'm working in that as well. I have to do some small naming changes and add a minimal test at least. So I might commit it by tomorrow.Am 10.04.2019 16:27 schrieb Pierre Bonneel <notifications@github.com>:No problem, I though that you erased the FEMM related commit by mistake. My bad. Take as long as you need ;)

PR #7, opened 2019-04-11
Bugfixes/lam holes

SebGue:
Hello Pierre,

my work on the LamHole related methods and functions has reached a reasonable state. 
Unittest runs without any errors and failures. However I found some bug, I will issue maybe tomorow.
Nevertheless you may merge my commit, so we are on par.

Best Regard,
Sebastian


BonneelP:
Hello Sebastian,

Thank you for your work. I've added few modifications and I will open some issues to keep in mind the improvements that we discussed. When I will have some time, my next work should be to add a validation case on the prius machine (IPMSM). 

Best Regard,
Pierre

PR #28, opened 2019-05-17
extended coil and turns computation

SebGue:


BonneelP:
Hello,

Great contribution, thank you :)
I have updated the error messages because Winding object are only in LamSlotWind object (for now)

Best regards,
Pierre

PR #29, opened 2019-05-21
 comp_connection_matrix() get Zs from slot by default 

SebGue:


BonneelP:

Issue #33, opened 2019-07-30
Rotor Initial Position


SebGue:
At the moment the initial electrical rotor angle related to the stator winding system for arbitrary pol/slot-cominations is not known.

It would be useful to either have a function that calculates the initial rotor angle, i.e. for example the angle between the d-axis and the alpha-axis, or to modifiy the FEMM model creation in a way that the d-axis match the alpha-axis (in case `angle_rotor_initial=0`).

BonneelP:
Solved in Pull request #34

Issue #26, opened 2019-05-06
get_surface_tooth


BonneelP:
For now we have a method in the Slot abstract object that returns the surface of the slot. It would be also interesting to have the possibility to define the surface of a tooth. For that it requires:
- To add a get_surface_tooth method in the LamSlot object that (for instance) returns the following surface (but centered on the X axis):
![image](https://user-images.githubusercontent.com/20533034/57218866-b755c400-6ff6-11e9-8573-4e6a78b1c516.png)
- Add a build_geometry_tooth method in the Slot object that calls the build_geometry method, apply some rotation, split some curves if needed, add the bore arc. The method would return the lines (in trigo order) 6,5,4,3,2,1,2,3,4,5,6
- The get_surface_tooth calls the build_geometry_tooth method and adds 3 lines to define the yoke part (7,8,7) and create the corresponding SurfLine object. The lines 7 are polar.

This method would enable: 
- To plot the tooth of the machine
- To get the surface corresponding to 1/ Zs of the machine
- This surface can be used to define a structural coupling (with gmsh for instance) by drawing a single tooth and then duplicate it by rotation.

We introduce the build_geometry_tooth method to be able to define some dedicated method for some complex slot if the general method doesn't work.

BonneelP:
Added in commit 7dcdc25049feaf5d860a6dee7b5b75e100a13ef8
Finally we have defined a "build_geometry_half_tooth" method that can be also used to compute the tooth width
Next steps: unittest for all Slot shape + remove bore line for Slot 60

BonneelP:
Validated in 32ea701

PR #34, opened 2019-07-30
stator winding voltage phasor angle calculation

SebGue:
This calculation may be needed to solve issue #33


BonneelP:
Hello,

Thank you for your contribution. Sorry for the late answer, I was in vacation. 
comp_initial_angle is indeed a very useful method. 

Regarding build_geometry_half_tooth, I think that it should be more a Slot method than a LamSlot. Indeed some slot may not be symmetrical and may need a redefinition of this method by inheritance. Why did you add this new method? Is it linked to Issue #26 ?

Great initiative to rename the label for FEMM for more consistence. 

Best regards,
Pierre

SebGue:
Hello,

I needed some method to calculate the width of the tooth. Therefore, I wrote the build_geometry_half_tooth method as a helper method, which also can be utilized to have the build_geometry_tooth method. 
Since the method uses the bore radius, I made it a LamSlot method. Otherwise one has to use "self.parent". But I don't know whats the better way.

Best regards,
Sebastian

BonneelP:
Hello,

Indeed placing this method in the Slot enforce the use of self.parent. But this constraint is already present in the Slots build_geometry methods (through the get_Rbo method). So I took the initiative to move the method to the Slot ones. 

I have also added a "is_top" optional parameter to select which half of the tooth to return and a test to plot the half tooths of a "Slot 10 machine". I won't add other test for the other slot, I will be tested through the get_surface_tooth method.

Best regards,
Pierre

PR #35, opened 2019-08-21
Winding Factor, Normalized Resistance, Conductor Thermal Coeff, ...

SebGue:
Winding Factor, Normalized Resistance, Conductor Thermal Coefficient, Matplotlib >= 3.3 Compatibility (lower case color keys)

### DIFF
--- a/Classes/MatElectrical.py
+++ b/Classes/MatElectrical.py
@@ -17,7 +17,7 @@ class MatElectrical(FrozenClass):
     # save method is available in all object
     save = save
 
-    def __init__(self, rho=1, epsr=1, init_dict=None):
+    def __init__(self, rho=1, epsr=1, alpha=1, init_dict=None):
         """Constructor of the class. Can be use in two ways :
         - __init__ (arg1 = 1, arg3 = 5) every parameters have name and default values
             for Matrix, None will initialise the property with an empty Matrix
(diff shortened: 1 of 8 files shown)


BonneelP:
Hello,

I have few minor remarks/questions: 
- According to the coding conventions  (http://pyleecan.org/coding.convention.html) I think we should renamed comp_norm_resistance into (for instance) comp_resistance_norm (for a better alphabetical sorting of the methods). What do you think is better, comp_resistance_norm or comp_res_norm ( with the idea to name all resistance related script comp_res_*) ?
- For comp_resistance_norm I thinking of adding in the docstring something like "only the active part, no endwinding" as a reminder for later work. Is it correct ?
- Great contribution to add comp_winding_factor, do you have the reference (book, publication...) from where you get your formula ?
- In comp_winding_factor, it is a great habit to check Zs type and >0 but thanks to the code generator, pyleecan raise an error if a user try to set an incorrect value for Zs. Although, what can happen is Zs = None.

Best regards,
Pierre

SebGue:
Hello Pierre,

you are right. We should rename the method to comp_resistance_norm. 
But I would like to avoid the abbreviation "res", since there many words starting like that, which may lead to some unnecessary confusion.
The method is thought to calculate the resistance per meter (not only for the active part). One could calculate the actual resistance by comp_resistance_norm * (length_active + length_end_winding). But you could also include your own factors e.g. for ac loss increase like that: comp_resistance_norm * (k_ac_loss*length_active + length_end_winding)
The winding factor calculation is based on complex phasor summation. I will have a look if i find some reference in english language. 
Good to know that the code generator raise an error. I wasn't aware of that.

Best regards,
Sebastian

BonneelP:
Hello Sebastian,

I also prefer to use comp_resistance_* there is less ambiguity. 

Thank you for your clarifications, as pyleecan will gather several models sometimes to compute the same quantities it is important that we always know exactly what each method is computing and how. 

By the way, the code generator also prevent to add new properties to the objects. For instance, it prevent some errors like setting (and creating) filename instead of setting file_name. 

Best regards,
Pierre

PR #41, opened 2019-09-04
[NF] Notches + Hallbach

SebGue:
This is the first proposal on Notch implementation
Therefore there are some short comings and limitations for now:
- notches on symmetrie lines are not possible, eg. on x-axis
- only the first notch (distribution) is used, ie. lamimation.notch[0]
- intersections eg. between different notch shapes are not checked

Possible future features
- notches on the yoke


BonneelP:
Hello Sebastian,

Thank you for your work ! 

Regarding your comment in create_FEMM_material: When there is no symmetry there are two surfaces that are need to draw the Lamination "Lamination_Stator_Bore" and "Lamination_Stator_Yoke". We only assign the material for the "Bore" line since it is the one we keep for symmetries. By the way, it will create some issue when splitting the surfaces. I thinking about introducing a new surface object defined by two surfaces. 

Great feature to add the possibility to split an Arc in FEMM.

For the intersection, I think that the key will be to use the method comp_angle_opening to know between which angles the notches are located. 

Best regards,
Pierre

PR #42, opened 2019-09-25
[NF] Slotless geometry with FEMM

Superomeg4:
Corresponding validation case: test_CEFC_001


Superomeg4:
Code formatted with black

PR #47, opened 2019-11-19
FEMM mesh & solution extraction

Superomeg4:
FEATURE INTRODUCTION:
This new feature allows to import the mesh nodes/connectivity/solution into the simulation and/or save these data in several .json file for latter use. 
To activate these features, the user have now 2 more attributes in the MagneticFEMM class: is_get_mesh and is_save_FEA.

NEW CLASSES INTRODUCTION:
A new "Mesh" abstract class is created. 
The new class MeshMat object is derived from Mesh. It defines a mesh which has its connectivity defined with a matrix (ndarray), because every element has the same type (Exemple: Triangle only).
The new class MeshFEMM object is derived from MeshMat. It has additional attributes specific to FEMM: the magnetic flux density B, the magnetic field H and the permeability mu, defined per element.
A new SubMesh class is also created (but it is not used yet): it contains a Mesh object which should be a smaller part of another Mesh object which is called the parent mesh.


BonneelP:

Issue #48, opened 2019-11-22
[NF] FEA mesh & solution class discussion 


SebGue:
Hello at all,

I want to start a disussion about a general FEA mesh & solution class here.
A first realization of this feature (more or less related to FEMM specifics) was done with pull request #47.
With the actual MeshMat class (which I think is meant to be a general class) it is only possible to store elements of one kind (i.e. triangles for example). I think this is very restrictive. 
Is there a reason not to expand the elements attribute ndarray to the max. number of nodes of the used or supported element types or to use a list to store the elements nodes. In addition this would require an element type attribute or to store the element type as the first value of each row within the elements attribute.

In the current implementation the FEMM solution is stored in a MeshFEMM object which is inherited from MeshMat. What do you think, if we seperate the mesh and the solution to some point, i.e. having a solution attribute in MeshMat. This attribue can then be of type FEASolution (or more general MeshSolution) with inherited FEMMSolution with nodal and elemental solution values.

At last, why are you interfacing FEMM with a LUA script instead of the python interface? Is it faster or more convienent for that purpose?

I'm looking forward for a constructive disussion.
Best regards,
Sebastian

Superomeg4:
Hello, 
Thank you very much for your constructive criticism, I really appreciate.

I am going to try my best to answer, but it is definitely an ongoing work which need to be improved.

1) About the restriction to ndarray:
As you see, the MeshMat is already inherited from an abstract class. The restriction to elements of the same type (triangles for instance) allows to build faster algorithms to use and postproc the solutions (by using vectorized matricial formulation of the equations). 
This can be definitely improved if there are mixed types of elements. I am open to any proposition, but I would like to keep the idea of ndarray containing all elements of the same type, which is really handy to use in postprocessing. 
Maybe we could think about your idea of storing everything with an "oversized" ndarray (with element type attributes etc.) , with additionnal functions which return a Submesh containing only elements of a certain type. What do you think ?

2) About FEMMSolution class:
I don't see any problem to do that, but it is important to keep a link between the mesh and the solution. At some point, the software could handle several mesh and several solution from different simulation. Maybe we should have a parent class "MeshSolution" with an attribute "Mesh" and an attribute "Solution". 

3) About LUA scripting:
It is much much more faster. It reduced the computation time to perform this operation from 5min to 10seconds in some of our cases. Actually, it's D. Meeker himself who gave us the trick.

Best regards,

Raphaël

SebGue:
[...]

PR #50, opened 2019-11-27
New architecture for Mesh module: Issue #48

Superomeg4:
The Mesh module was reorganized taking into account the discussion in Issue #48. The goal is to be more flexible, such that it allows multiple ways for the "user" to store the simulation results.

NEW CLASSES INTRODUCTION:
1) MeshSolution: New class which must contain one Mesh object and one Solution object.

2) Mesh: Must contain an Element object, a Node object, and a list of mesh stored in the attribute submesh. (SubMesh object was removed).

3) Element: New class which must have a all necessary "get" methods. The way to store attributes (connectivity ...) is defined in the inherited class. This give more freedom to create new way for storing mesh, as long as all the getter are properly defined. 

4) ElementDict and ElementMat: Both inherited from Element. ElementMat is using the previous way of storing elements (i.e. only one type of element, using matrix). ElementDict is using dictionnaries to be able to store several types of elements. 

For example, one could write an additional "ElementGmsh", which would mimic the way gmsh works.

5) Node & NodeMat: same principle as for Element but applied to node storage. 

6)  Solution : Object to store data related to the solution, it must have a "get_field" method. SolutionFEMM is specifically defined for FEMM.


BonneelP:
Thank you for your contribution Raphaël !

SebGue:
Hello Raphaël, hello Pierre,

I'm not quite sure, if it make sense to have ElementMat- and ElementDict class.
I think that for both classes the getter methods are class specific. 
For example the get_connectivity method will produce the following results:
`elements.get_connectivity(elem_type="Triangle")`
On an instance of ElementDict this may return a result, while an instance of ElementMat will raise an error.
`elements.get_connectivity()`
Will work on both (ElementMat and ElementDict), but will only return an (random) portion of the ElementDict-mesh.

Further, if you'd extend get_connectivity with a parameter elem_type on ElementMat class, you also have to introduce an attribute element_type.
And I also think that the number of nodes won't be sufficient to decide about the element type. 
Think for instance of Tetrahedron and Quadrangle elements. Both got 4 nodes. Or think of higher order elements.

My proposal would be to set the element attribute of Mesh class to be a dict. ElementMat class will then be the values of this dict and have a default list element types as keys. The element order can then be checked by the number of nodes.
This also allows for additional user element classes ("ElementGMSH" for example), as long as the pyleecan list of default element types and the getter specification are regarded.
That way also a lot of your code can be reused with only slight modifications.

What do you think? 

Best regards,
Sebastian

Superomeg4:
Hello Sebastian, 

Thank you for your feedbacks !
You are perfectly right, the methods are not "pure" enough because the output depends too much on the class type. 

So if I understand correctly, Mesh should have an attribute from the class ElementDict() which would contain a dict() with object from inherited the class Element() classes (for example ElementMat or ElementGmsh). Such that ElementDict is no more inherited from Element. Is that right ? 

Then what should be the correct behaviour for get_connectivity ? 

Best regards,

Raphaël

SebGue:
Hello Raphaël,

I think the main point is, that get_connectivity (and all similar methods) should have the element type as parameter for convenience. 
No matter if ElementDict class is a dict populated with ElementMat instances or if ElementDict stay as is. I'm fine with both solutions.
ElementDict will be inherited from Element in any case (as opposed to my initial proposal of Mesh's attribute "element" to be of class dict, which doesn't make sense with object orientation and user abilty to extend pyleecan).

Do you think that could be a possible solution for you too?

Best regards,
Sebastian

Superomeg4:
Hello Sebastian,

I have discussed with Pierre, and we have a nice proposition:
- The Mesh will now have an attribute element which is a dict() containing only Element. This way, one can easily loop on all element type to perform adapted operations. mesh.element("Triangle") will contain an Element inherited class (which can be named "Triangle") storing all elements relative to 3 nodes triangles. Thus we store everything by element type.
- The getters (such as get_connectivity) will be moved into the Mesh class. This makes more sense because we juggle between nodes and elements. 
- The method get_connectivity will be split into 2 method: get_all_connectivity(type_elem = "Triangle") which return a matrix of connectivity and a vector with the corresponding element tags number, and get_connectivity(tag_elem = number_of_the_target_elem) which can return only one element at a time. Following the same logic, I may change other methods.

If someone wants to store a mesh in a completely different way, he could define another Mesh class with similar getters (but I think it will be better to have adapted "Import" methods).

What do you think ?

SebGue:
Hello Raphaël,

I'm fine with your proposal. So the different element types can be accessed in a convienent way.

I will put a link to this discussion into #48 for completeness and also some other related topics, later on.

PR #51, opened 2019-12-04
[WiP] Additionnal features and methods for Mesh module : Issue #48

Superomeg4:
Some bug fixes
New features to add elements to an existing ElementDict object
New features to compute the interface between two meshes (Example with a mesh of Segments at the interface of two meshes of Triangles)


Methods/Mesh/Mesh/interface.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
+from pyleecan.Classes.ElementDict import ElementDict
+from pyleecan.Classes.NodeMat import NodeMat
+from collections import Counter
+import numpy as np
+
+
+def interface(self, other_mesh):
```
Superomeg4: Maye we should split this method in several function.

Superomeg4:
Adding the node_tag attribute for NodeMat classes. Otherwise it is going to be to complicated to manage submeshes.

Superomeg4:
Modifications from discussion #48 have been taken into account.

Classes/Mesh.py
```diff
@@ -61,14 +182,43 @@ def __init__(self, element=None, node=None, submesh=list(), init_dict=None):
                 submesh = init_dict["submesh"]
         # Initialisation by argument
         self.parent = None
-        # element can be None, a Element object or a dict
-        if isinstance(element, dict):
-            self.element = Element(init_dict=element)
+        # element can be None or a list of Element object
```
BonneelP: a list => a dict

BonneelP: a list => a dict

Classes/Mesh.py
```diff
@@ -61,14 +182,43 @@ def __init__(self, element=None, node=None, submesh=list(), init_dict=None):
...
+                            + " in init_dict for element"
+                        )
+                    # Dynamic import to call the correct constructor
+                    module = __import__(
+                        "pyleecan.Classes." + class_name, fromlist=[class_name]
+                    )
+                    class_obj = getattr(module, class_name)
+                    self.element[key] = class_obj(init_dict=obj)
+                else:
+                    element = element  # Should raise an error
```
BonneelP: else:
    self.element[key] = obj

Classes/Mesh.py
```diff
@@ -97,10 +247,18 @@ def __str__(self):
             Mesh_str += "parent = None " + linesep
         else:
             Mesh_str += "parent = " + str(type(self.parent)) + " object" + linesep
-        if self.element is not None:
-            Mesh_str += "element = " + str(self.element.as_dict()) + linesep + linesep
-        else:
-            Mesh_str += "element = None" + linesep + linesep
+        if len(self.element) == 0:
+            Mesh_str += "element = []"
```
BonneelP:
```suggestion
            Mesh_str += "element = {}"
```

BonneelP:

PR #61, opened 2020-01-01
Optional extendable GUI

SebGue:
Hello,

I've added the option to use an user extendable GUI. 
Therefore I also did some changes to the Machine.plot method to also plot some intermediate steps.
There is also a slight BC that verifies that the loaded 'json' is an Machine instance.

Best regards,
Sebastian


BonneelP:
I particularly like how you reorganized the machine plot.

BonneelP:
As a side note, it would be interesting to check how the "plot" part of the extended GUI behave at each step of the machine GUI (starting from a new machine) since for now most of the parameter are set to None at the beginning of the process. Or we can go forward on the "defaut value" approach for the GUI

SebGue:
Actual the plot of a new machine will cause an error (TypeError 'None'). I think thats okay for now.
This also led me to the question, what is the purpose of the check methods. Maybe they could be utilized to check if a specific machine part can be plotted.
"Default values" are a nice to have feature and i think we should keep that in mind. But in my opinion there are some other issue with a higher priority, i.e. Material Library (I will open an issue.) and iron losses.

BonneelP:
Indeed the check method enables us to know which part of the current machine are correctly set. We use them to know which step of the nav_step are available. So we can use it to know which part of the machine we can plot.

PR #65, opened 2020-01-15
[NF] load/save of data structures (of lists and dicts)

SebGue:


Functions/save.py
```diff
@@ -9,7 +9,70 @@
 from os.path import join, basename, isdir
 
 
-def save(self, save_path=""):
+def fix_file_name(save_path, obj):
+    if isdir(save_path) or not save_path:
+        file_path = join(save_path, type(obj).__name__ + ".json")
+    elif "." not in basename(save_path):
+        file_path = save_path + ".json"
+    else:
```
BonneelP: For now it checks that there is a "." in the path but not that it ends by ".json"

BonneelP:

Issue #58, opened 2019-12-18
Automated GUI - Proof of Concept


SebGue:
Hello to everyone,

this issue will be to discuss the possiblites to automate the GUI creation as mentioned [here](https://github.com/Eomys/pyleecan/issues/17#issuecomment-566557388).

> Since there are a lot of hand coded files, what do you think about an object attribute editor instead? I don't actually know if that will work as expected, but it would track user class modifications easily without the need to adapt the GUI.
>
>> Why not. I'm just not sure how it would work with the current GUI. The structure of the GUI enables to easily add new slots, magnets, conductor... which is mostly the new objects that contributor should add. But as we have the csv doc, I think that we should be able to generate automatically such GUI. So it is worth investigating.

What keeps me promising is the following blog post. http://trevorius.com/scrapbook/python/attribute-editor-in-pyqt/

@BonneelP: Could you tell me why `machine.__dict__` return underscored attributes as keys, `_rotor` for instance, although I can access the rotor with `machine.rotor` (without underscore)? 

Best regards,
Sebastian

BonneelP:
Hello Sebastian,

The "_property" is linked to the class generator. The last part of the class file the getter and setter of the class. When accessing to a property we store them into "_property" instead of "property" to make sure that the getter is called. You can use "machine.as_dict( )" instead of "machine.__dict__" to get all the properties.

But I think that the csv documentation will be more helpful for the generation since we have the property and there corresponding type which would enable us to adapt the widget to the property.

Best regards,
Pierre

PR #67, opened 2020-01-16
 GUI LamHole mag_mat select for every magnet

SebGue:
Added w_mat_select widgets to select the magnet material of every magnet of a LamHole lamination.
Also 'blacked' the GUI files which is maybe not necessary.


SebGue:
Hello Pierre,
I think, I'm finished here for now. 
Best regrads, Sebastian

BonneelP:
I didn't merge it before because of the conflict. Thank you for your last commit !

BonneelP:
As a further evolution, what about having the material setup directly in the "PHoldM5X" widgets ? When combining several kind of holes or several layers of the same hole we may want to define different materials. Do you think that we should investigate the possibility to set a magnet to "None" ?

SebGue:
There are several things coming together here and I will try to address them but only in short:
Direct material setup would be good, since the material setup is already per hole. But that's not clear for the user at the moment.
Therefore a more "automated" GUI would be great IMO (cf #58 ).
I think we already disussed "None" magnets in the past, but I didn't checked yet. Nevertheless I like the idea.
I think hole magnets are a little inconsistent at the moment since they are no "real" magnets, i.e. they are no MagnetTypeXY and have different of methods (get_height_magnets vs. comp_height).
If I remember correctly multiple holes are not implemented till now. Only the first hole is regarded.

SebGue:
One more thing: It would be good to have the possiblity to include a magnet gap to include manufacturing tolerances for instance. Maybe it is sufficent if the gap will be "virtual", i.e. by modifing the material parameters of the magnets.

BonneelP:
To be more clear on the schematics we should add a label with the magnet number. 

I just realized that everything is called "slot" on the widget related to "holes", should we rename it ?

If a magnet of a HoleM5X is set to None, the build_geometry will remove the corresponding surface which is quite interesting.

Indeed for now all magnets are rectangular and defined by the hole itself. Maybe for later development we can say that the hole define the area dedicated to drawing the magnet and the Magnet object would change how it is defined.

It is already possible to define several holes with the GUI with the button "add new slot". That's why having the material setup in SMHoleMag is tricky if there are several kind of holes with a different number of magnets.

Issue #60, opened 2019-12-25
Class Generator: Taking into account ndarray stored in dict


Superomeg4:
Hello everyone,

I have encountered a limitation to the actual class generator. The method "save" output is not JSON serializable if some ndarray are stored into a dict. 

When an attribute is directly a ndarray, the method "save" turns it into an list (which is JSON serializable). But if this ndarray is stored into a dict, the method "save" does not perform this conversion.

Best regards,

Raphaël

SebGue:
Hello everyone,

since the following issue seems to be related to this issue, I will also place it here:

It would be great if the "save" method would be able to save python default types, i.e. especially dicts and lists. That way it will be possible to have "collections" of pyleecan data, for instance machine or material databases or multiple simulation outputs within one file.

Best regards,
Sebastian

BonneelP:
Hello everyone and best wishes for the new year !

Regarding Raphaël's issue: The method to work on is not "save" but "as_dict". It is working for pyleecan object containing ndarray because in the "as_dict" method we convert the ndarray to list (and in the __init__ method we convert it back to ndarray with "set_array"). I see three solutions: 
- We introduce a new type for the class generator "dict(ndarray)" that can contains only ndarray and that do the correct convertion in as_dict and __init__
- We don't use dictionary of ndarray but dictionary of Pyleecan object containing ndarray
- We extend the save method to another way to save a python object which is not json

Regarding Sebastian's issue:
save is a method of any pyleecan object (thanks to the class generator). To be able to save a list or dict of pyleecan type we would need to create a "save_list" or "save_dict" function and extend the load function to take this new possibility into account. These new functions shouldn't be that hard to create since we can reuse the "as_dict" method.

Best regards,
Pierre

SebGue:
Hello Raphaël, hello Pierre,

@Superomeg4 could you please post some code to reproduce your issue.

I think the issue could be related to the class generator in general (not only the as_dict method).
I would expect to be only permitted to set the defined type of dict elements. But with the following code it is possible to also set ndarray dict elements for instance.

```
from pyleecan.Classes.Mesh import Mesh
from pyleecan.Classes.NodeMat import NodeMat
from pyleecan.Classes.ElementMat import ElementMat
import numpy as np

mesh = Mesh()
mesh.element["Triangle3"] = ElementMat(nb_node_per_element=3)
mesh.element["ndarray"] = np.array([1, 2, 3])

```

[...]

PR #71, opened 2020-01-29
Save/Load dict of ndarray in .json : Issue #60

Superomeg4:
This new feature introduce a new type {ndarray} for the class generator, which correspond to a dict of only ndarray.

Then the generated classes are able to correctly save/load from .json file objects containing dict of ndarray.

If a dict must contain several type of python object, including ndarray, the user should create a new adequate classe and use the feature about dict of pyleecan objects. 

Thus, this pull-request should close the issue #60 .


Superomeg4:
I wanted to modify the Solution class, and it was why i created this issue initially. Thus, I have modified the test_CEFC_002_save_mag to test the new feature for dict of ndarray.

BonneelP:

PR #74, opened 2020-02-10
[NF] InCurrentDQ

SebGue:
This is a proposal of an InCurrentDQ class that can be used to input dq-reference frame stator currents instead of phase currents with InCurrent class.
I copied and adapted the InCurrent class code. I think this is the most straight forward solution.
An other solution could be to include an optional current transformation within InCurrent class.
What do you think?

Best regards,
Sebastian

### DIFF
--- a/Classes/InCurrentDQ.py
+++ b/Classes/InCurrentDQ.py
@@ -1,0 +1,524 @@
+# -*- coding: utf-8 -*-
+"""File generated according to pyleecan/Generator/ClassesRef/Simulation/InCurrentDQ.csv
+WARNING! All changes made in this file will be lost!
+"""
+
+from os import linesep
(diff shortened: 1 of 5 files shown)


BonneelP:
Hello Sebastian,

I think that it is a great idea ! It will be more convenient to use pyleecan with this feature.

I don't know what is better between having several InCurrent classes or lot's of option in the InCurrent class. 
With the first solution, one can pick the object that fit its data and everything is straightforward but that may creates lot's of Input object. 
The second solution will only keep the main Input objects (which is better to think workflow wise) but each object will be more complicated with options and parameters that are not always needed. 

So for now I think that we can keep your object and see how the usage evolve in the future (the input code is already a bit complex with all the verifications!)

Anyway, thank you for your contribution!
Best regards,
Pierre

PR #76, opened 2020-02-14
[BC] InCurrentDQ -> load_switch.py

SebGue:


BonneelP:
Hello Sebastian,

I agree with all your modifications. I'm just wandering why are you saving the class_dict into json ? Do you plan to use it in another part of pyleecan?

Best regards,
Pierre

SebGue:
Hello Pierre,

indeed I would like to use the class_dict.json for at least a tree diagram of pyleecan classes.

Best regrads,
Sebastian

BonneelP:
Hello Sebastian,

Great idea! 
On this topic, we were thinking about using dot language to automatize the generation of the UML diagram from the class_dict. In the Architecture documentation we used dot to generate the UML images. Here is for instance the code that was used for the machine and lamination part: 
[machine_and_lam.txt](https://github.com/Eomys/pyleecan/files/4218110/machine_and_lam.txt)
This code is shouldn't be too hard to generate according to the class_dict. We also used this tool to preview the results: http://graphs.grevian.org/graph

Best regards,
Pierre

PR #80, opened 2020-02-24
[NF] Reorganisation of class generator

cedric-marinel:
[NF] The class generator handle imported types from external packages
[NF] The class generator handle function as property
[CC] generate_class function is now split in several packages
[NF] Pyleecan objects __str__ methods call recursively __str__ methods instead of calling as_dict


BonneelP:

PR #81, opened 2020-02-26
[BC] Correct tests bugs caused by class generator modifications

CedMrnl:
[BC] Corrects bugs caused by the last PR


BonneelP:

PR #82, opened 2020-02-27
Optimization

CedMrnl:
Add a global optimization tool based on [DEAP](https://github.com/DEAP/deap) using [NSGA-II](https://www.iitk.ac.in/kangal/Deb_NSGA-II.pdf) algorithm.

The tool consider only minimization problem, handle constraints and allow to define many types of design variables.


BonneelP:
Thank you for this great new feature :)

PR #83, opened 2020-02-28
[NF] Plot mesh from FEA solving with pyleecan colors with plot_mesh method

Superomeg4:
Validation case with test_CEFC_002_save_mag


BonneelP:

PR #85, opened 2020-03-05
[NF] relative paths + [BC] InCurrentDQ

SebGue:
- added missing function of InCurrentDQ
- material data from material library use relative path now


BonneelP:
Great idea to gather everything in definition. Maybe one day we can extend it to other global variables (Exceptions ? plot colors ?)

PR #86, opened 2020-03-10
Logging + Optimization improvements

CedMrnl:
[NF] Add a generic way to define loggers in Pyleecan.

All logger are defined in _Pyleecan/loggers.py_ in a dict config inspired from [this example].(https://stackoverflow.com/a/59241472).

A default logger _Pyleecan_ is defined and specific loggers are defined for the following classes:
- Output
- Simulation
- Machine
- OutElec
- OutMag
- OutStruct
- OutGeo
- OptiGenAlg

The classes above have in their csv file the _logger_name_ attribute corresponding to a logger defined in the config dict in _Pyleecan/loggers.py_.
Two config dict have been defined to use console text loggings or files. The config choice is in _Pyleecan/\_\_init\_\_.py_.

Every class has a method get_logger that get the class logger or the parent class logger or the default logger.


BonneelP:

PR #89, opened 2020-04-10
Use of SciDataTool for Br and Bt plots

helene-t:
[NF] New plot commands are now available, which rely on the SciDataTool package.

Simulation data (for the moment Br and Bt) are stored into Data objects, and can be extracted in slices, Fourier Transformed, converted, compared, etc.

To be extended to other simulation fields, declare it as a DataND object in Classesref, and create the objects at the end of your simulation module.

Available plot methods:
- plot_A_space
- plot_A_time
- plot_A_surf
- plot_A_time_space
- plot_A_cfft2
- plot_A_fft2
- plot_A_nthoct


BonneelP:

PR #90, opened 2020-04-16
Relative imports

### LINKED ISSUE #79: [CO] Change of ROOT_DIR
Hello Pierre,

what do you think about a change of the "root_dir" of pyleecan to ..\XYZ\pyleecan instead of ..\XYZ ?
Therefore all imports have to be corrected, but on the otherhand its enable the user to rename the pyleecan folder without breaking the program.

Best regards, Sebastian

CedMrnl:
[NF] replace absolute import by relative imports to fix #79
[CC] Remove headers to avoid copyrights  on pyleecan website

Before imports in every pyleecan file were:
```python
from pyleecan.Classes.Circle import Circle
```

Now it has to be:
```python
from ....Classes.Circle import Circle
```

Theses changes allows to rename pyleecan folder. The GUI has to be regenerated to work in the new folder.


BonneelP:

PR #91, opened 2020-04-22
Relative imports

CedMrnl:
[NF] Change the repository organization according to #79 to be able to install pyleecan. 
Now, it can be installed with the following command:
`python setup.py install`

Once pyleecan is installed, the GUI can be lauched with the command:
`python -m pyleecan`

The relative imports are now only in the pyleecan folder.


BonneelP:
Thank you very much for this important step :)

PR #98, opened 2020-05-14
Force module

Superomeg4:
This PR proposes to separate the Force calculation from the Structural module. 
As a consequence, the Output object contains a new OutForce object. 

Additionally, the class ForceMT now returns SciDataTool objects.


BonneelP:

PR #100, opened 2020-06-05
[WP] GUI option widget

CedMrnl:
A default user folder is created to store the data and an option widget has been add in the GUI to handle different material libraries


BonneelP:

PR #106, opened 2020-06-16
New syntax for plot commands using SciDataTool

helene-t:
The get_along methods of SciDataTool now return a dict containing the axes and the field.

The plot commands have been modified accordingly.


BonneelP:

Issue #62, opened 2020-01-07
[BC] Material Library


SebGue:
Hello to everybody,

I have been testing the current implementation of the MatLib-Editor within the pyleecan GUI and recognized some short comings. For instance, if a material is edited, the changes are not propagated thoughout the whole machine.
Therefore I would like to discuss the behavior that everybody would expect from the Material Library Editor.
Here are some thoughts and questions from my side.
- changes to material library should propagate to all machine parts if they use the respective material, i.e. material object references should not been broken
- Material Library should show the origin of the currently selected material (MatLib or Path)
- it should not been allowed to have duplicate material names
- if a loaded machine contains "new" material, the user should be asked to add this material to the library, to keep it as is or to replace it with a material from the library that has the same data, if it exists
- if there are materials with the same name (in the machine file and/or the library) the user should be asked to rename the material
- an alternative may also be to keep the machine files materials "outside" the library and only inside the material selection combobox with the possiblity to also edit that material (that may require some additional logic to handle all possible cases)

All of your contribution to this issue is greatly appreciated.

Best regrads,
Sebastian

BonneelP:
Hello Sebastian,

Thank you for this interesting issue. The reason the Matlib is working that way is a legacy from MANATEE where the material are stored in separated files from the machine one. For MANATEE only the name of the Material is meaningful for the selection and everything is loaded at the beginning of the simulation. The MatLib only update the material file for later use.

I agree that we need to change that to have something more coherent with pyleecan. 

Note that we currently save the material within the machine object which can cause the following issue: I create a machine with Mat_A, save it, change some value of Mat_A latter, then come back to the old machine which still has the old value of Mat_A (but I may expect to have the new value). To avoid that when loading a machine we can overwrite the material with the current version of the material library. If we need to create such function when loading a machine we can call this function when updating the material library. That way the material in the MatLib are the reference.

The GUI breaks the material reference to enable the user to cancel its modification which can be interesting when you may change a material used in several machines. But it may be not necessary if we create a copy of the material and then call the init method of the current instance with the as_dict of the copy. 

Regarding the "origin" of the material I don't understand what you mean but we can display the path to the file.

I agree for the duplicated name

> if a loaded machine contains "new" material, the user should be asked to add this material to the library, to keep it as is or to replace it with a material from the library that has the same data, if it exists

I agree if we detect a new material by checking only its name. 

What do you think of the idea to use the MatLib as the reference for all material handled within the GUI ? 

Best regards,
Pierre

PR #110, opened 2020-06-19
[BC] Solve some small bugs

CedMrnl:
[WP] Changes to material library now propagates to all machine materials


SebGue:
Hello Cedric,

I know the material library is 'Work in Progress', I still got some things to mention.
- Maybe you could document, what the MatLib is intended to do especially e.g. if there are new materials within a machine file or if material names already exists.
- In the MatLib dialog it is necessary to click twice to select a material if one switches between MatLib materials and machines materials.
- If a material of a loaded machine already is in the library although the paths are different, the paths are not updated.
- The two material selection lists within the dialog are a little confusing to me. Maybe you could name them something like "Library Materials" and "Machine Materials (not contained in the library)". It also would be really nice to see what's the difference between two materials of the same name, maybe with the tooltip feature.

Some time ago, I introduced the 'Functions/path_tools' to have relative paths for the materials path properties. Due to the recent changes of the imports, this feature doesn't work anymore. Maybe it is also not necessary with your changes to the material library, so it could be removed. You could discuss it. If you want to keep it, I would suggest to replace the indirect import with a direct import, i.e. 'from ..definition import MATLIB'.

Best regards, 
Sebastian

CedMrnl:
Hello Sebastian,

I was going to post a message in #62  to discuss with you about these changes. 

I created a MatLib object to keep the reference to the Material Library everywhere in the GUI. Before, the material library was only contained in a list and we could lost the reference by replacing the material library in a widget by a new list. 

MatLib contains:

- list_mat: list containing the materials from both the material library and the machine
- index_first_mat_mach : index of the first machine material in the list, or len(list_mat) if there is no machine specific materials
- machine: current machine to edit its materials once a material is edited

"MatLib" class name is maybe not really clear as it stores all the materials including the machine ones. I can rename it in GuiMaterials, what do you think ?

I have modified the Material folder location. By default the folder pyleecan/Data is copied in a new Pyleecan folder depending on the user OS. This folder also contains a config_dict to store user preferences. This behaviour is defined in _pyleecan/definition.py_ and avoid to delete user materials once we will install pyleecan with pip. The GUI takes the folder Pyleecan/Data/Material as default library but a new material library folder can be loaded in GUIOption widget. 

For now, when we load a new machine we compare its materials with the MatLib excluding material name and path. If a material is not found in the MatLib we add it in the machine specific materials part of the list. If the material is found we replace the machine material by the material found in the machine. 

If two different materials have the same name, we rename the last one by adding "(1)".

When a machine is saved, its materials are also saved in the same file. This behaviour avoids to modify  a machine that has been used in a simulation when we modify materials in the material library and is in favor of reproducible science.

With this behaviour material's attribute path can be confusing so we should remove it and only save the material according to its name and the current material library path defined in GUIOptions. 

I will add some modifications to DMatLib to be clearer and to allow to put material from material library into machine materials.

Do you see any trouble that could arise with this way of managing materials? 

Best regards,
Cédric

SebGue:
Hello Cédric,

so the "MatLib" class is essentially an interface to a specific folder (containing json material data files). Regarding the class name, I don't care.
The only thing that is also important to me is reproducible simulations. Some time ago I already had a discussion with Pierre about that. At the moment, to have reproducible simulations, we have to think about the imported material data, e.g. the "ImportMatrixXls" class that only stores the path of the data.

Further, is it possible to have multiple machines have access to the "MatLib" class? The reason is, I was thinking about some GUI modifications to have more than one machine loaded into the GUI.

Otherwise I don't see any trouble at the moment.

Best regards, Sebastian

CedMrnl:
Hello Sebastian,

MatLib class enables to separate materials from material libray (contained in a specific folder) and machine-specific materials specific (contained in the machine file). For the moment, every material is stored in the same list but I will split them in a dictionnary to handle several machines.

I agree with you about the "ImportMatrixXls" class. As a quick fix we can define the B-H curve in the GUI as a:
```Python
ImportMatrix(ImportMatrixXls('file.xls').value)
```

Best regards,
Cédric

CedMrnl:
Hello Sebastian, 

I pushed some modifications. Now MatLib contains a dictionnary which splits reference material library ("RefMatLib" key) from machine specific materials ("MachineMatLib" key). I think this organization should enable you to handle several machines with their own materials in the GUI in a cleaner way. 

I also fixed some bugs according to materials. If a material is edited in the DMatLib from the SidebarWindow, changes propagate to all machine parts.

The Machine's attribute _path_ is now a relative path from config_dict["MATLIB_DIR"].

Do you think that issue #62 can be closed? 

Best regards,
Cédric

BonneelP:

PR #112, opened 2020-06-19
[NF] Torque in a SciDataTool.DataND object for fft and plot commands

helene-t:
- Torque (Tem) is now stored in a SciDataTool.DataND object so that out.plot_A_time("mag.Tem") may be called.
- New methods to compute average between yoke and slot radii


BonneelP:
Tank you for your contribution :)

PR #118, opened 2020-07-01
[WP] Save and load files in hdf5 format (.h5)

CedMrnl:
Pyleecan object can be stored in hdf5 files. However, the method can be improved because SciDataTool objects are serialized using cloudpickle before being stored.


BonneelP:

PR #121, opened 2020-07-08
Code cleaning in definitions.py

IrakozeFD:
Code cleaning inside definitions file
- default_config_dict removed from definitions to default_variable file
- generic functions created in init_environment file to initialize pyleecan environment


BonneelP:

PR #123, opened 2020-07-09
Curve colors and colormap from config file

helene-t:
Colors for curves and colormaps are now imported from the config file, and can still be replaced by the user for each plot command.


BonneelP:

PR #130, opened 2020-07-21
[BC] Correct reference to machine materials in the MatLib

CedMrnl:


BonneelP:

Issue #128, opened 2020-07-16
[NF] Multi-simulation organization


CedMrnl:
Hello all, 

I would like to add some classes to define and compute multi-simulation.
I just wrote a tutorial to explain how it would be organize on [pyleecan-doc](https://github.com/Eomys/pyleecan-doc/blob/master/tuto_multisimulation.rst). It describes new classes : 

- *VarParam* to define the multi-simulation,
- *ParamSetter* to set the parameters,
- *DataKeeper* to define the results to keep after each simulation,
- *XOutput* to store results. 

This new organization would involve few modifications in Optimization part: a new generation would be defined as a multisimulation and OutputMultiOpti would be modified to heritate from XOutput.

I think this organization would enable to define any kind of multi-simulation, what do you think ?

Best regards,
Cédric

PR #131, opened 2020-07-24
[WP] Multi-simulation workflow

CedMrnl:
This PR introduces a first organization described in #128 to define multiple simulations based on:
- a reference simulation, 
- parameter values to explore with *ParamExplorerSet* class,
- output data to keep with *DataKeeper*.  

In order to harmonize multi-simulation and optimization workflow, optimization has to be changed with *ParamExplorer* as design variables and *DataKeeper* as objective functions to minimize. 

*OptiSolver* returns now a *XOutput* object so *MultiOutput* class will be deleted soon. Some regressions have been added as *plot_pareto*, *plot_generation*, *plot_generation_design_space* and *plot_pareto_design_space* are not included yet in *XOutput*.

Cédric


BonneelP:

PR #132, opened 2020-07-27
Added function to generate 2D geometries in Gmsh format.

ajpina:
Hi Team,
An added function to create 2D geometries/mesh in Gmsh format with gmsh api. The function is fed by build_geometry() method in stator, rotor and shaft objects. Duplicated lines and points are removed and a new dictionary is created with the actual gmsh tags for lines and points. 
Magnets and windings can be removed with the options is_lam_only_S/R. 
Overlapping points/lines from build_geometry() are not handled thus far.
Test function is also added.

### DIFF
--- a/Tests/Functions/test_gmsh2.py
+++ b/Tests/Functions/test_gmsh2.py
@@ -1,0 +1,47 @@
+import pytest
+
+from os import makedirs
+from os.path import join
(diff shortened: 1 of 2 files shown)


BonneelP:
Thank you for your contribution :) This feature will be very useful !

PR #137, opened 2020-08-04
[WP] Optimization modifications according to multi-simulation organization

CedMrnl:
Edit optimization objects to match better with the multi-simulation organisation (#131) 

*OptiProblem* now takes in argument list of *OptiDesignVar* (*ParamExplorer* daughter), *OptiConstraint* and *OptiObjFunc*. 
Results are stored in a XOutput where the *DataKeeper* contain fitness values and *ParamExplorerSet* contain design variables values.

*XOutput* has two optimization plot plot_generation and plot_pareto inspired by *OutputMulti* previous implementation.


BonneelP:

PR #141, opened 2020-08-19
[NF] Electrical module

helene-t:
[NF] Compute voltage and torque from Id/Iq or Ud/Uq

Electrical torque (Tem_av_ref) is now computed in the electrical module. One can either impose current or voltage, however the computed inductance is not correct if Id/Iq are not provided.

A simple electrical equivalent circuit is used. Others can be easily added in the future.


BonneelP:

PR #146, opened 2020-09-04
Updates for MeshSolution module

Superomeg4:
Introduce a new Interpolation feature for FEA post-processing purpose. 
Interpolation object can be stored as attribute of CellMat.
Interpolation objects contains 3 attributes:
- GaussPoint objects (for quadrature purpose)
- RefCell (to define the type of cell and interpolation order). Example: Triangle3 or SegmentP1 are provided.
- ScalarProduct (to define the type of scalar product which can be used to perform integrations and variational formulation calculation).

+ various CO and BC.


BonneelP:

PR #148, opened 2020-09-07
Refactoring unittest to pytest

BenjaminGabet:
Refactoring all unittest to pytest : 

- All "import unittest" have been removed;

- Every assert.Equal, assert.True/False, assert.AlmostEqual have been replaced;

- Most of the tests are now in test classes;

- Some marks have been created : METHODS, PLOT


CedMrnl:
I realized that some scripts were executed during the test collection e.g.: 
```python
SCIM_001 = load(join(DATA_DIR, "Machine", "SCIM_001.json")) # <--- This line is executed during the collection

def test_desc_SCIM():
    """Check that the description of a SCIM is correct"""
    desc_dict = SCIM_001.comp_desc_dict()
    assert len(desc_dict) == 9
    assert desc_dict[0]["name"] == "Type"
```

All lines not defined in a function are executed during the test collection even if all tests in the file are deselected.
To avoid this behaviour we should define machine in the test itself or in a fixture if the machine is used in several tests:
```python
@pytest.fixture(scope="module")
def SCIM_001(): # <--- This function is executed just before the test
    machine = load(join(DATA_DIR, "Machine", "SCIM_001.json")) 
    return machine

def test_desc_SCIM(SCIM_001):
    """Check that the description of a SCIM is correct"""
    desc_dict = SCIM_001.comp_desc_dict()
    assert len(desc_dict) == 9
    assert desc_dict[0]["name"] == "Type"
```

We could also use [indirect parametrization](https://docs.pytest.org/en/stable/example/parametrize.html#indirect-parametrization) to avoid the parametrization list to be executed during the collection. 

Best regards,
Cédric

BonneelP:

Issue #127, opened 2020-07-15
[NF] Class Generator - List of ndarray


SebGue:
Hello Pierre,

it seems there is no type '[ndarray]' implemented yet. Maybe you could add this feature easily. (For me, it would take some time, since I'm not into the ClassGenerator.)

Best Regards, Sebastian

BonneelP:
Hello Sebastian,

For which application do you need a list of array ? Is it possible to add another dimension to your array or to use a Data object ?

Best regards,
Pierre

SebGue:
Hello Pierre,

it is for an application I told you before. (I have to discuss if we publish it if I'm done with it.)
But that may also apply to the 'MachineUD' class, i.e. I want to store multiple Br, Bt results corresponding to multiple lamination gaps for example. 
For now I use pure 'list' type. BTW: Is it possible to save lists of ndarrays?

OT: With my new classes, I see a lot of possible improvements especially to the FEMM integration. I will discuss all my findings with you in a while.

Best regards, Sebastian

helene-t:
Hello Sebastian,

We are currently working on a new SciDataTool object: VectorField will be an object to store fields with 2 or 3 components (e.g. magnetic flux B will contain Br and Bt). Components will be DataTime or DataFreq objects.
VectorField will have `get` methods similar to DataTime or DataFreq objects, with access to radial, tangential and axial components (e.g. `get_rad_along("time", "angle=0")` or `get_mag_tan_along("time=0", "wavenumber=[0,80]")`). We plan to replace Br and Bt by such an object in OutMag. We will let you know as soon as those changes are effective.

Concerning multiple results, you may want to use an additional axis in a SciDataTool object:
```
Airgap_pos = Data1D(
    name="airgap",
    values=positions_array,
)
Br = DataTime(
    axes=[Time, Angle, Lam],
    values=Br_array,
)
Br.get_along("time", "angle", "airgap")
```

Best regards,
Helene

SebGue:
Hello to all,

could anybody help me, since I'm struggling setting up a list of SciDataTool.Classes.DataND.DataND type in the ClassRef.
Or is this not possible at the moment.

BR Sebastian

CedMrnl:
Hello Sebastian,

This feature is not available yet. Maybe we could improve the class generator to handle dict and list of external type, or just SciDataTool types. 
[...]

PR #150, opened 2020-09-14
[NF] Class generator improvements

CedMrnl:
Class generator now enables to:

- define any Pyleecan object as an attribute using *FrozenClass*.

- define list of external type using *[SciDataTool.Classes.Data1D.Data1D]*

The second feature should solve #127


BonneelP:

PR #155, opened 2020-09-18
Change in multi-simulation (VarSimu) and FEMM computation (MagFEMM)

CedMrnl:
VarSimu always compute the reference simulation before other simulation to enable to use its results for the others.

MagFEMM can now import a *.fem* file.


SebGue:
@BonneelP Hey Pierre, do you know why you introduced _FEMMHandler instead of straight forward 'import femm'?

PR #161, opened 2020-09-24
[NF] Post class for postprocessings

CedMrnl:
Add a _Post_ class to define postprocessing in _Simu_ and _VarSimu_.


BonneelP:

Issue #156, opened 2020-09-21
Optimization Module


SebGue:
Hello Eomys,

I had the chance to finally try the optimization module, i.e. the optimization tutorial. 
@CedMrnl Thank you for implementing this great feature!

My questions are:

1. Is there actually the possibility to have dependent quantities? E.g. if I would like to have a constant air gap and my design variable is the height of the (surface) magnets, some other geo. parameter had to be adapted to maintain the air gap width.

2. Can I reuse the results of a previous optimization simulation for a 'new' simulation with bigger population or a higher number of generations?

Best regrads, Sebastian

CedMrnl:
Hello Sebastian, 

Thank you for your feedback!

For your first question, I think you can use the `OptiDesignVar.setter` by defining a defining a function that takes a simulation and a value in argument. 

For instance, if you want to modify the size of all your slot parameters, you can define the following function and `OptiDesignVar`:

```python
def slot_scale(simu, scale_factor):
   """Edit stator slot size according to a percentage

   Parameters
   ----------
   simu: Simulation
     simulation to modify
   scale_factor: float
     stator slot scale factor
   """
   simu.machine.stator.slot.W0 *= scale_factor
   simu.machine.stator.slot.W1 *= scale_factor
   simu.machine.stator.slot.W2 *= scale_factor
   simu.machine.stator.slot.H0 *= scale_factor
   simu.machine.stator.slot.H1 *= scale_factor


OptiDesignVar(
    name="Stator slot scale factor",
    symbol = "S_sf",,
    type_var="interval",
    space=[
        50,
        150,
    ],
    get_value=lambda space: random.uniform(*space), # To initiate randomly the first generation
    setter=slot_scale,
```

For your second question, the feature is not available yet be it would be great to have it. 

Best regards,
Cédric

SebGue:
Hello Cédric,

that seems to be exactly what I need :-)
I missed that setter can be a function, since in the tutorial it is a string.
So for OptiDesignVar class the setter function must have 2 inputs. The first input passed by OptiDesignVar is always the coressponding simulation and the second input is the design variables value.

The attribute design_var of OptiProblem class is a list. What is the execution order of this list. Is it from first list element to last list element?
Then it would be possible to have a last 'unused' design variable (e.g. type = 'set', space = [ 1 ]) with a setter, that is doing all the work?! This should not affect optimization itself.

[...]

PR #162, opened 2020-09-25
[NF] Optimization improvements

CedMrnl:
This PR add some features to the optimization module discussed in #156:
- add _OptiProblem.preprocessing_ to run a preprocessing before running the simulation
- add _OptiProblem.datakeeper\_list_ to save some output data for every individual
- delete _OptiGenAlgNsga2Deap.toolbox_ to avoid loading issue


BonneelP:

PR #163, opened 2020-09-28
[CC] fixed SyntaxWarnings + some typos

SebGue:


SebGue:
I also had some changes due to 'black' (I had v19 installed.) before I updated to v20.
Maybe we should define an 'official' black version to use to avoid commit noise.

BonneelP:
Hello Sebastian, 

Thank you for noticing :)
I also agree that we should use an "official" black version. Is there a particular method to select which one or should we all use the latest one (with a check when running the class generator) ?

Best regards,
Pierre

SebGue:
Hello Pierre,

I think we should do both, choose a version (which might not always be the latest) and check when running class generator with a warning as output.

Best regards,
Sebastian

BonneelP:
Hello Sebastian,

I have set the official black version as "20" and added the proper warnings in 341a17f46e3c4f6604a33469f11258c76a40bdaa

Best regards,
Pierre

PR #170, opened 2020-09-30
Various optimization and correction for FEMM

EmileDvs:
[BC] Include rotor initial position in rotor angle to align rotor d axis and stator alpha axis

[BC] Debug calculation of rot_dir depending on winding configuration
[CO] Put park transform in RMS value to match with pyleecan standard
[CO] Use of symmetries in MagFEMM model using comp_sym() of machine object
[CO] Put Phi_wind_stator in data object
[CO] Put default minimum angle as 15° instead of 30° to reduce number of elements


SebGue (commented):
Hello @EmileDvs,

you did some changes regarding the transformation and the rotation direction, I don't understand.
Could you give me some further explanation on that.
You also caluclate the RMS value with the n2ab transformation while the inverse transformation (ab2n) doesn't, if I got it right. So ab2n(n2ab(Z)) would result in an other value than Z.
It would be good if you at least add the conversion from amplitude to rms to the doc string.

Best regards, Sebastian

EmileDvs:
Hello @SebGue,

Thanks for your feedback! Don't worry we are still working on this PR, the fact is we have noticed some conflicts with Pyleecan conventions: 

- motor convention, i.e positive torque when rotor rotates in increasing angle, which requires that rotation direction of fundamental magnetic field is rot_dir=-1, such that B_fund = B*cos(2*pi*freq*time + rot_dir*p*theta) translates to increasing angle when time increases. The n-phases should then be dephased of rot_dir*2*pi/n to be in-line with Pyleecan convention. Then if winding phases are reversed, rot_dir changes of sign and inputs currents are reversed as well.

- rms convention, i.e Id_ref and Iq_ref are specified as rms but when generating instantaneous currents in ABC frame they are divided by sqrt(2). As you mentioned it, this is due to the fact that Park Transform conserves quantities as rms. Therefore, the PR agrees with what we expect from dq2n transform in Pyleecan: 

       1) Get Id/Iq in rms from any set of 3-phases current in function of time
       2) Get the set of 3-phases current function of time for a given Id/Iq in rms

Since this no longer Park transform, we can add some booleans in the functions, for example, is_dq_rms and is_n_rms that we respectively set to True and False for Pyleecan current convention and that we can also use to get Park transform.

I hope this a bit clearer now, let us now if not :)

Best regards,
Emile

BonneelP:
By working on the coupling with FEMM we realized that the symmetry wasn't automatically applied in FEMM (it was still parameters from Magnetics). We are introducing a new organization on how we handle "periodicity" (which is a better name). Now Time and Angle axes are defined with the expected periodicities from the Input object (in gen_input => comp_axes). Then in MagFEMM, the periodicity is extracted from the axes. A new parameter "is_periodicity" (replacing all is_sym_X) in Magnetics enables to disable the periodicity. Should we have is_periodicity_t and is_periodicity_a ? At the end we save only the periodic part, which will significantly reduce the memory use. The aim is that all module can use the periodicity through SciDataTool by calling the getter for full or periodic Data.

The code is still messy for now and we need to do some modification on SciDataTool to really handle the periodicity. More information/discussion incoming.

BonneelP:
Hello all,

There are still few tests to check and correct but the modifications are and work on most of the Tests. So the aim was to fully introduce the periodicity (previously called "symmetries") in FEMM. Previously there was only space periodicity in FEMM and it was set manually and the full matrix was stored. 

- periodicity_t and periodicity_a as property of Magnetic to activate both time and space periodicity.
- Time and Angle are now Data object to be used as reference for the module (instead of ndarray). They are set in Input.comp_axes (for all Input/physics).
- Periodicity of the machine is now gathered in OutGeo with a method Output.get_machine_periodicity (to compute it only once).
- Na_tot and Nt_tot were removed from OutMag. Now one need to call get_length(is_oneperiod) of an axe to know the full or the periodic length. This enable to always know what we are dealing with in the computation.
- solve_FEMM now use both time and angular periodicity and store only the periodic part. get_values of Data object return by default the full matrix to avoid problem with model that can't handle periodicity
- ForceMT can reuse the periodicity of B. 
- rot_dir was duplicate in OutElec and OutGeo. Now it is in OutGeo only (with a proper getter)
 
All these modifications are done to improve the preformation in computation time and memory of pyleecan. 
This PR use the next release of SciDataTool that can be installed with:
`pip install --index-url https://test.pypi.org/simple/ SciDataTool`

Best regards,
Pierre

Issue #154, opened 2020-09-17
Results plots don't show off


SebGue:
Hello everybody,

I tried to plot the results of a simulation (in IDLE) as it is done in the tutorials, e.g. with `myResults.plot_A_space('force.P', is_fft=True, r_max=40)`, but the plot don't show off. 
What am I doing wrong? Or is there some 'fig.show()' missing in the plot code?

Best regards, Sebastian

CedMrnl:
Hello Sebastian, 

You're right, polt_A_space method does not return the figure and does not call `fig.show()`. I think that every plot should call automatically `fig.show()`.
As a quick fix in your script, I think that you can use `plt.show()` from *matplotlib.pyplot* or you can get the figure with `fig = plt.gcf()`.

Best regards,
Cédric

PR #171, opened 2020-10-01
[BC] show figure of plot_A_xxx methods

SebGue:
Automatically show figures if no 'fig' input parameter is provided. See #154 
In general there is some inconsistancy between the plot_A_xxx methods regarding 'fig' input parameter. So it's maybe worth to revise them.


BonneelP:

Issue #165, opened 2020-09-29
[NF] Defining the EEC for the electrical module for squirrel cage induction motor


MVreemann:
I have been trying to get pyleecan to simulate a regular squirrel cage induction motor (SCIM) but i cant get it to work. I will attach my code to this post as well as the motor i defined using the GUI and a material .json file for Aluminium i made. This is just a fairly standard SCIM which i am trying to analyze for a project i am working on.

My code is basically a copy of the code found under tutorials on the main website. I know there is a key difference between both motors because there my motor doesn't have any permanent magnets in the rotor. But i reckoned i should at least get some form of output. Since there still is an input for the rotor current (albeit a total guess).

However i do not understand the error i got:

```
  File "......\pyleecan-master\scripttest.py", line 51, in <module>
    out = simu.run()

  File "......\pyleecan-master\pyleecan\Methods\Simulation\Simulation\run.py", line 43, in run
    self.run_single()

  File "......\pyleecan-master\pyleecan\Methods\Simulation\Simu1\run_single.py", line 19, in run_single
    output.geo = self.machine.comp_output_geo()

  File "......\pyleecan-master\pyleecan\Methods\Machine\Machine\comp_output_geo.py", line 28, in comp_output_geo
    output.rotor = self.rotor.comp_output_geo()

  File "......\pyleecan-master\pyleecan\Methods\Machine\LamSlotWind\comp_output_geo.py", line 21, in comp_output_geo
    output = Lamination.comp_output_geo(self)

  File "......\pyleecan-master\pyleecan\Methods\Machine\Lamination\comp_output_geo.py", line 22, in comp_output_geo
    output.name_phase = self.get_name_phase()

  File "......\pyleecan-master\pyleecan\Methods\Machine\LamSlotWind\get_name_phase.py", line 22, in get_name_phase
    return gen_name(self.winding.qs)

  File "......\pyleecan-master\pyleecan\Functions\Winding\gen_phase_list.py", line 75, in gen_name
    for i in range(N):

TypeError: 'NoneType' object cannot be interpreted as an integer
```
What i get from this is that the name of the phase can not be determined because there is no element to name? I have to admit my python skills have deteriorated over the last few years and weren't all that good to begin with. So the answer might be really simple, but i just can't seem to figure it out.

Does anyone understand the problem?

[Files.zip](https://github.com/Eomys/pyleecan/files/5297482/Files.zip)

BonneelP:
Hello,

[...]

PR #175, opened 2020-10-05
[NF] EEC_SCIM

SebGue:
Hello,

this is the first working (but only slightly tested) version of an EEC_SCIM class. #165
Nevertheless it has lot of limitations and TODO's yet.
I.e. the machine eec parameters have to been manually input to the parameters dict. 
Also the eec is only voltage driven for now. Therefore the Input has to be InputElec.

Best regards, Sebastian


BonneelP:
Hello Sebastian,

Thank you for this PR. Should I merge it now or do you prefer to work on some "TODO" before ?

Great idea to introduce felec instead of slip, it is more generic. Do you think that you can introduce a "set_felec(slip)" as we have a "set_Id_Iq(I, Phi)" ?

Best regards,
Pierre

SebGue:
Hello Pierre,

you can merge this PR, since I don't know how much time I can spent on that issue in near future.
But before, could you add the "set_felec(slip)" as a TODO somewhere in the code?! Thank you!

BR, Sebastian

PR #178, opened 2020-10-07
[BC] Electrical FEMM simulations

SebGue:
This PR resolves a serious issue with the FEMM calculations in the Electrical module.
The issue was, that the Input could have an other number of rotations than the actual FEMM simulation.
Therefore the currents were calculated (transformed) incorrectly, leading to wrong results.
Further the inital rotor angle was disregarded.

TODO: The electrical tutorial has to been updated. (One can see the wrong results there.) Since I haven't used the notebooks yet, it would be good if anyone else could do that.

Best regards, Sebastian


BonneelP:
Hello,

Thank you for this modification :) 
We also corrected few strange behaviour with the rotor_angle in the FEMM branch. 

Great that you merged the flux_linkage computation into a function. I sure we can go further on this idea.

Regarding the tutorials, all the latest modifications on our side are architecture optimization according to our latest experimentation with pyleecan. We plan to validate and merge all the pending PR before the first webinar. All the tutorials will then be updated and will be used during the webinar on a release 1.0.1. We will then focus on v1.1 to use it in the last webinar if possible. 

Best regards,
Pierre

Issue #173, opened 2020-10-02
GUI migration from PyQt5 to PySide2


IrakozeFD:
Hello Everyone, 

I open this issue to talk about the PYLEECAN GUI which is currently developed with [PyQt5](https://riverbankcomputing.com/software/pyqt/intro) (from Riverbank). 
My concern is about the license behind PyQt5:  GPL. The [GPL License](https://www.gnu.org/licenses/licenses.en.html) is not quite permissive if you want to develop 
a software without distributing the source. The issue here is that you will have to purchase a commercial license from Riverbank 
for PyQt5 to use PYLEECAN GUI in your private software. 

As workaround I propose to migrate the GUI code to [PySide2](https://www.qt.io/qt-for-python) (From Qt) which is under [LGPL license](https://www.gnu.org/licenses/lgpl-3.0.en.html). With this license it could be possible to use 
PYLEECAN and its GUI in a private software, you only need to ensure that the code covered by the LGPL (here the GUI code) is made available including modifications made to it.

Thank you

PR #179, opened 2020-10-07
GUI with PySide2

IrakozeFD:
This PR migrates the PYLEECAN GUI code from PyQt5 to PySide2, the main objective is to prevent license issues that could occur when one wants to use PYLEECAN GUI in a private software. In fact, the license behind PySide2 is LGPL, which enables the usage of the code under LGPL without distributing the entire source code of the project  (on the condition that this code under LGPL is made available including modifications made to it). 

Fixes Eomys/pyleecan#173

Best regards, 

Franco


BonneelP:

PR #182, opened 2020-10-11
[BC][CO] Optimization

SebGue:
[CO] removed limitation, that optimization simulation output could only be of type Output
[BC] adapted plot_generation to recent changes in output of Optimization


BonneelP:
Thank you for noticing, it should have been changed when we introduced XOutput.

Best regards,
Pierre

PR #183, opened 2020-10-11
[BC] ClassGen, copy method

SebGue:
Had some strange optimization results ... took me hours to find the bug :-(
I thought `type(obj)(init_dict=obj.as_dict())` would deep copy a pyleecan object (e.g. used in copy method), but it doesn't.
At least if the object contains a dict.
This PR fixes this behavior as long as the dict has only mutable types in it. So it should only be a temporary workaround.
I think we should further investigate this issue also for lists. A possible solution could be a function that walk though all list/dict elements and serialize them as well.
Maybe we could also have a TestClass, i.e. a class specifically for unit testing only, that has all possible combinations of property types in it, to avoid such issues in future.

Best regrads, Sebastian

BTW: This PR has #182 in it. I hope this is not a problem.

### DIFF
--- a/pyleecan/Classes/DXFImport.py
+++ b/pyleecan/Classes/DXFImport.py
@@ -116,7 +116,9 @@ class DXFImport(FrozenClass):
 
         DXFImport_dict = dict()
         DXFImport_dict["file_path"] = self.file_path
-        DXFImport_dict["surf_dict"] = self.surf_dict
+        DXFImport_dict["surf_dict"] = (
+            self.surf_dict.copy() if self.surf_dict is not None else None
+        )
         DXFImport_dict["BC_list"] = self.BC_list
         # The class name is added to the dict for deserialisation purpose
         DXFImport_dict["__class__"] = "DXFImport"
--- a/pyleecan/Classes/DataKeeper.py
+++ b/pyleecan/Classes/DataKeeper.py
@@ -138,14 +138,20 @@ class DataKeeper(FrozenClass):
         DataKeeper_dict["name"] = self.name
         DataKeeper_dict["symbol"] = self.symbol
         DataKeeper_dict["unit"] = self.unit
-        if self._keeper_str is not None:
-            DataKeeper_dict["keeper"] = self._keeper_str
-        else:
+        if self.keeper is None:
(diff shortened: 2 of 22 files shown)


BonneelP:
Hello,

Thank you for noticing. I discussed about it with Cedric and he propose that we use [copy.deepcopy](https://docs.python.org/fr/3/library/copy.html) for dictionary and list. 
Indeed we can add a "_AllGenPossiblePropTest" class and it should be tested alongside all the others. 
Should we merge this PR and open an issue to remind us about the other necessary modifications ? Is there a way so that we can contribute to your PR (other than modifying the code from Github) ? 

Best regards,
Pierre

SebGue:
Hello Pierre,

you can merge this PR. 
Is copy.deepcopy able to copy pyleecan objects as well?

> Is there a way so that we can contribute to your PR (other than modifying the code from Github) ?

I don't know exactly what you mean or what you want to do. If you want to modify my PR before merge, you could pull my ClassGen branch into a new branch on your repo and I close this PR. But I guess you already know.

Best regards, Sebastian

SebGue:
... or do you think of something like that: [Invite Collaborators](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-your-github-user-account/inviting-collaborators-to-a-personal-repository)

BonneelP:
Hello,

Yes I was looking for a solution like "invite collaborators" but without the invitation ^^' Anyway, I will merge this PR and do the suggested modifications on PR FEMM (it is fully validated now and will be merged soon. It will be the ref for release 1.0.1)

Thank you again for this contribution :)
Best regards,
Pierre

SebGue:
I send you  an invitation though. Just to test this feature. I will also add an issue where I will collect some thoughts uptill tomorrow.
Should I also invite someone else?

BonneelP:
Thank you for the invitation :) We can talk about other invitation tomorrow.
The commit above introduces the use of copy for list as well. In as_dict for list and dict of both ndarray and pyleecan object we already create a new dict and list and call as_dict or tolist (return a copy of the array) for each element. So as long as the dict and list don't contain object it should be Ok (and in that case we should use [obj] for instance and we should be able to detect it).

PR #184, opened 2020-10-12
[CC] lambda function strings

SebGue:
This PR updates some lambda functions to lambda function strings (according to #172).
It also add the possibility to use np and random in the function string, e.g. `func_str = "lambda x, np=np: np.abs(x)"`
(This seems a little odd, but will work for now. Maybe someone has a better solution.)

### DIFF
--- a/Tests/Plot/test_ICEM_2020.py
+++ b/Tests/Plot/test_ICEM_2020.py
@@ -612,13 +612,11 @@ def test_Optimization_problem():
     # -------------------- #
 
     # Objective functions
-    def tem_av(output):
-        """Return the average torque opposite (opposite to be maximized)"""
-        return -abs(output.mag.Tem_av)
+    """Return the average torque opposite (opposite to be maximized)"""
+    tem_av = "lambda output: -abs(output.mag.Tem_av)"
 
-    def Tem_rip_pp(output):
-        """Return the torque ripple """
-        return abs(output.mag.Tem_rip_pp)
+    """Return the torque ripple """
+    Tem_rip_pp = "lambda output: abs(output.mag.Tem_rip_pp)"
 
     my_objs = [
         DataKeeper(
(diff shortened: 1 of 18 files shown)


BonneelP:
Hello,

I run some check on my side, as long as the package is imported there is no need to provide it as argument of the lambda.
I will update the class generator to always import numpy as np and random when a function is provided. A test including np. will be added in test_post_var_simu

Best regards,
Pierre

SebGue:
That's really strange. I get the 'random' of 'np' not definded error???

BonneelP:
Strange indeed :/ I noticed that some of your lambda already didn't have np. as argument. Maybe it is linked to some lambda ? Normally eval create the lambda in the context of the class file which have the corresponding package so we shouldn't have to provide np as argument. I just pushed the corresponding modification, I will run all the tests to make sure.

SebGue:
I have double checked. I also thought it should be sufficient to have np/random imported at lambda definition, but it was not.
I set a breakpoint in at lambda definiton in 'def _set_get_value' of OptiDesignVar to check the import, i.e. `dir()` showed that random was imported. Nevertheless exactly this design var resulted in the mentioned error.
BTW: With your latest commit, do you import random and np at every class?

BonneelP:
On my side all the tests are Ok. Which version of python are you using ? I'm on 3.8.6
In the latest commit I import random and np for all the class that have at least one property of type "func"

SebGue:
Oh... I haven't seen the 'if func' :-)
I am on 3.8.3
Maybe it will work with the 'global' class import. I will try.
Have you checked the 'if 'np.' in value' import as well?

BonneelP:
We I first tested it before my commit, I was already using "global" class import so that can be the explanation for the difference :)

SebGue:
That's it. With global class import the error is fixed. 
So I quiet have to learn a lot regarding python :-)

BonneelP:
Great :) I think that luck help me a bit on this one ;)
I have push an improvement for #183 as well. I'm running all the tests one last time if it is Ok I will merge #170, then update the tutorials and create the release 1.0.1 for the webinar.

PR #190, opened 2020-10-20
Femm

EmileDvs:
The aim of this pull request is to add FEMM parallelization on time step by calling several FEMM instances ("workers") in parallel using multiprocessing. MagFEMM object has now one additional property "nb_worker" to specify the number of workers to use. Two methods are added : solve_FEMM_parallel and get_meshsolution_parallel. There is one validation case "test_FEMM_parallelization" to validate the feature.

The second aim is to validate the MagFEMM object with the webinar tutorial 01 and several validation cases : 

- test_FEMM_periodicity
- test_EEC_FEMM_IPMSM
- test_EM_FEMM_IPMSM_varload

The PR also correct minor bugs, such as : 

- limiting M400-50A BH to H=170 kA/m, after which incremental permeability is mu0 (before incremental permeability was going under mu0). 

- always open FEMM as background task (especially when calling FEMM to calculate EEC parameters)

- save FEMM models used for EEC parameter calculation in the Results folder


EmileDvs:
In addition to the two original purposes, this pull request brings two major code optimizations to better account for periodicity and use SciDatatool to plot Data object:

- the old plot_A_space/time/fft methods and functions have been merged into the generic plot_2D_Data functions and methods, calling plot_2D (renaming of plot_A_2D). Thanks to SciDataTool flexibility, the plot call is now transparent since the axes to be plotted are directly put as arguments, using the same syntax as calling SciDataTool object through get_along. This makes easier the use of normalization (for example, convert wavenumber into space orders or frequencies into electrical orders), the boundaries of each axis, and the superposition of several results for comparison.  All the validation cases and tutorials have been updated to account for this new changes, and can serve as example to perform plots. Furthermore, a lots of example will be also put in the plot tutorial, especially for next webinar. The last update

- the axes (in particular Time and Angle axes) are now passed from one physics to another using comp_time_space method. This method handles machine (anti-)periodicity, simulation periodicty (if they are activated or not for ex) and also computes additional axes used in the model calculation. For example, the comp_time_space method of Magnetics calculates Time and Angle axes on which to calculate airgap flux density. Using polymorphism,  the method comp_time_space  of MagFEMM (which inherits of Magnetics) first call the  comp_time_space  of Magnetics to get Angle and Time axes, and then add an additional Time axis for torque calculation. In fact Time axis for torque can be different from the one used for flux calculation since if flux is (anti)-periodic then Time axis is periodic. Furthermore, the comp_time_angle of Force module changes axes in Magnetics module to remove anti-periodicity is any because forces are no-longer anti-periodic.

A few other changes and corrections are also put into this pull request:
- Due to the last update of SciDataTool, all symmetries are now stored only in axes (no longer in DataTime/DataFreq/VectorField objects) to avoid redundancies
- Automatic calculation of fundamental harmonic to color it in black in FFT 1D plots, accounting for several normalizations: wavenumber, frequency, spatial order, electrical order, mechanical order etc. and several quantities: flux density, currents, maxwell force density, stator winding flux.
- Simplification of unit-mmf calculation and plot: the mmf_unit is no longer stored in the output and so there is no need to check if it exists before loading it or calculating it. In fact, it is very fast to compute so there is no real need to store it. mmf_unit is only useful to get rotation direction of magnetic field fundamental, stator d-axis position and to plot winding function and unit mmf and is recomputed to perform any of these tasks.

Please tell me if you have any questions!

BonneelP:
I've run all the tests on this branch alone and it is all clear but there are some bugs on master. I have corrected everything in PR #201. Once this new PR merged we should be able to do the release 1.0.2 for the webinar.

Note that this PR use the new release of SciDataTool 1.1.5 (requirements have been updated)

PR #194, opened 2020-10-22
[CO] extended search for find_cell + minor changes

SebGue:
This PR propose an extended search for find_cell method (get the cell that contains a given point) of MeshMat. 

With the actual implementation there were cases, were no cell (of type triangle) was found, e.g. with the following values:
```
# Triangle 
(x1, y1) = 0, 0.5
(x2, y2) = 1, 0.1
(x3, y3) = -1, 0.1
# Point
(x, y) = 0, 0
```
Although the point is nearest to node 1, no cell is found, since only cells that contain node 1 were searched.
Now, after a failed first search, all other cells are tested if they contain the point (ordered by their cell center).

### DIFF
--- a/pyleecan/Classes/CellMat.py
+++ b/pyleecan/Classes/CellMat.py
@@ -101,7 +101,7 @@ class CellMat(FrozenClass):
         nb_cell=0,
         nb_pt_per_cell=0,
         indice=[],
-        interpolation=None,
+        interpolation=-1,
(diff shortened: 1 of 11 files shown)


Superomeg4:
Good job ! This was a major issue I haven't though about. 
The PR is OK for me, but we definitely should add a test_find_cell method.

BonneelP:

PR #196, opened 2020-10-23
enable SCIM simulation with rotor currents

SebGue:
This PR mainly enables simulation of SCIM (LamSquirrelCage rotors) with rotor currents.
The rotor bars now have only positive winding direction, with one phase each bar.

There are also some bug fixes and code optimizations included.
At most the names/labels of the surfaces of the windings are unified to contain 'Stator' or 'Rotor' instead 'S' or 'R' by the introduction of the Slot class method get_name_lam. This also enables to easily change the names e.g. to 'Lam_x' later on (with MachineUD).

### DIFF
--- a/Tests/Functions/test_gmsh.py
+++ b/Tests/Functions/test_gmsh.py
@@ -13,6 +13,7 @@ from Tests import save_validation_path as save_path
 
 
 @pytest.mark.GMSH
+@pytest.mark.long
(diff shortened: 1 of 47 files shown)


BonneelP:
Thank you for this PR. Some label clean up won't hurt the project and being really able to set current on the rotor as well is a great step.

PR #199, opened 2020-10-25
Modified function to add air gap sliding band for 2D geometries in Gm…

ajpina:
A new parameter is added to draw_GMSH() function, the parameter is_sliding_band=[True,False] allows the code to add four layers in the air gap so the mesh can be used in a FE solver for electromagnetics such as Elmer. The four layer approach will enhance the results regarding to forces calculation in the air gap. The number of circumferential elements in the air gap can be set by user_mesh_dict={} parameter as shown in the example in the test file attached. Two layers will rotate with rotor while the other two stay fixed with the stator.
I will work next in the boundary conditions to be added in the gmsh file and ease the settings for solving.


BonneelP:
Hello, 
Thank you for this new improvement for the coupling. We have started to investigate the coupling for GMSH, we will see if we can further improve it. If you have some idea on how, we will be happy to help you!

It seems that there are some conflict with your PR so I can't merge it now, can you take a look to both files ?

Best regards,
Pierre

ajpina:
Hi Pierre,
It should be solved now.
Best,
Alejandro

BonneelP:

By the way, we have updated the class generator this week: now when you call as_dict on a class containing a complex they are stored as str. Then when loading they are converted back to complex (from instance cf https://github.com/Eomys/pyleecan/blob/master/pyleecan/Classes/Arc1.py, line 349 and line 302). It may help for Tests/Functions/test_gmsh2.py line 114.

Best regards,
Pierre

PR #203, opened 2020-10-28
Post plot

EmileDvs:
The aim of this branch is to add a new class PostPlot which can be used to automatically execute plot as postprocessing of the simulation.


EmileDvs:
The PostPlot definition is very close to the plot method to call. Required arguments must be put in param_list while optional arguments must be put in param_dict, such that the call to the plot method is:
`plot_method(
        *self.param_list,
        **self.param_dict,
    )`
The PostPlot automatically saves the figure in the results folder. The user can specify a different folder by specifying "save_path"=path_str or not save the figure by specifying "save_path"=None. 

This branch also enables to pass fig and ax as argument of the plot function/method so that it is possible to make user-defined subplots. The test file is pyleecan\Tests\Plot\test_PostPlot.py. In this test, 2 subplots are plotted, each using 2 PostPlot objects:

-  the first subplot contains radial and tangential flux densities PostPlots, 

- the second subplot contains the machine PostPlot alongside torque versus time PostPlot. 

A fifth PostPlot is executed in background to illustrate how to save to svg.

The direct use of subplots can also be found in the plot function/method plot_A_time_space, which has been updated to benefit from the new plot_2D/3D_Data syntax.

BonneelP:
Thank you for this new feature :)

PR #204, opened 2020-11-01
MagElmer Skeleton

SebGue:
I added MagElmer class with some methods as a skeleton for magnetic Elmer simulation. Hope it is helpful somehow.


SebGue:
@BonneelP I did some change to OutMag, I would like to be approved by you. Thx.

BonneelP:
Hello, 

It make sense to rename FEMM_dict to FEA_dict indeed :) 
The skeleton is Ok for me. We can adapt it if the Elmer workflow requires it. We will see if some global function can be moved to Magnetics. I merge this PR as MagElmer is a separate object that won't affect the others. Maybe for collaborative structural PR we can open a new branch one day.

Best regards,
Pierre

Issue #200, opened 2020-10-27
[CO] Clean import and saving of Data objects in MagFEMM module


EmileDvs:
Hello,

This issue deals with improving current import from electrical output inside the FEMM model and put all savings of Data objects (B, Tem, Phi_wind_stator etc.) on a higher level (preferably in run() method of Magnetics)

At the moment currents are set in set_FEMM_circuit_prop by doing "I.values" which is not a robust way to get values from Data object, because it implies that time values in electrical output are exactly the same as those in the magnetic model. Besides all physical quantities are saved in solve_FEMM and solve_FEMM_parallel which duplicates code that could also be useful for other magnetic models (for example, calculate bemf or torque ripple).

This issue can be an important step to demonstrate how the run() method of a physical model can be organized:

1. First define axes to store in the Output and to  (here in MagFEMM, compute Time and Angle using comp_time_angle method). This method can be renamed "comp_axes_dict" for more genericity. axes_dict contains axes that are used for calculations
2. Import output from previous physics (here currents from electrical output) on the axes calculated in first step
3. Run calculations (here comp_flux_airgap)
4. Save outputs in Data Objects (all should be put in run())

Let me know if you have any feedback on this.

Emile

SebGue:
Hello Emile, 
I think I got some addition. Since MagFEMM is steady state only, BEMF for one current only differ by time/rotor angle axes. So it would be good to have the possibility to 'recalculate' for different speeds without the actual FEMM calculation.
Therefor it may be a good idea to have a is_steady_state property in Magnetics class to determin if a recalculation is suitable.

Best regards, Sebastian

EmileDvs:
Hello Sebastian,

Thanks for your feedback.

I am going to work on this pull request for the last Webinar, to have a clean demonstration on how a physical model in Pyleecan can be implemented. Besides, we now have a new Magnetic model (MagElmer) so we'll make sure that all of these changes work well for both FEMM and Elmer models. For your information, I am going to do it on the FEMM branch of EOMYS-Public.

Concerning your point, I completely agree to add the solver type to the Magnetics class since it is compulsory for flexible solvers such as Elmer. Regarding FEMM, FEMM can solve both magnetostatic and magnetoharmonic problems. Even if we only have static solver at the moment, the solver type is also relevant for FEMM. 

[...]

PR #206, opened 2020-11-02
[CO] Reorganize Magnetics and MagFEMM input and outputs

EmileDvs:
This PR aims at solving issue https://github.com/Eomys/pyleecan/issues/200

The PR is work in progress, I'll update it as things progress.


EmileDvs:
The Magnetics and MagFEMM objects are now updated with the changes detailed in https://github.com/Eomys/pyleecan/issues/200

The standards outputs of any Magnetics models are stored in the end of Magnetics.run(), e.g. B, Tem, Phi_wind_stator data objects.

Outputs that are only related to MagFEMM are stored in the end of MagFEMM.comp_airgap_flux(), e.g. the MeshSolution containing mesh information and magnetic quantities per elements. To do so I had to improve the FEMM parallelization, in particular reuse solve_FEMM in solve_FEMM_single function of solve_FEMM_parallel. FEMM parallelization can be further improved following new issue https://github.com/Eomys/pyleecan/issues/208.

Sebastian, Alejandro, I think we should update the MagElmer architecture as well, in particular removing the part that store standards outputs because it will be done in Magnetics.run(). Please let me know if you need any help on this!

SebGue:
Hello Emile, is this PR ready to merge or is it still wip?

SebGue:
Hello @EmileDvs, hello @BonneelP ,

since you are working on solve_FEMM, I ask my question here.
Could there be class specific code in solve_FEMM, e.g. `if isinstance(winding, WindingSC)`? I tend to say no. But maybe you got other opinion.
Background is, that actually periodicity (a) with induction machines doesn't work, i.e. periodicity is calculated and set correct for the model but the expected number of phases (by solve_FEMM to get the circuit props) doesn't match. So there is an error.

BR, Sebastian

EmileDvs:
Hello @SebGue,

I agree with you, what I did is not clean at all. As you pointed out, the space periodicity do not apply on rotor phases of SCIM. What we want is to specify to comp_I_mag only the rotor phases (i.e. the rotor bars ID) that will be drawn in the model and for which currents must be calculated. In conclusion, this is all about taking into account periodicities at the beginning of each model, and we can do as we did for time and angle. My new proposition is to store the phase axes (both stator and rotor) in the axes_dict that we pass to comp_flux_airgap.

Still, in comp_time_angle we still have to check whether spatial periodicity contained in Angle axis apply to stator/rotor phase axes, i.e. to check in the winding is a "WindingSC". Because we don't want to apply it the other windings types.

While you may come up with a better proposition than mine, I am modifying the pull request to implement the phase axis in comp_time_angle (that we should probably rename comp_axes, as it is already the case in gen_input).

Best regards,
Emile

SebGue:
Hello @EmileDvs ,

that's fine to me. Just to conclude: You will do the 'general' reorganization and I will add the special case.
My idea is to use the Laminations comp_per. method to calculate the models qs (number of phases) and reconstruct the other phases props. afterwards (utilizing 'wind_mat').
BR, Sebastian

EmileDvs:
Hello @SebGue,

Ok then the most simple is not to implement periodicity on phase on the present pull request. We can open an issue later to solve this problem for Squirrel Cage Induction Machines specifically.

Best regards,
Emile

EmileDvs:
After discussion with @BonneelP  and @helene-t, I have created the issue related to the phase axes in https://github.com/Eomys/pyleecan/issues/211. 

We also decided not to merge this PR before the next webinar in order to improve a bit further the way magnetics data coming from MagFEMM.comp_flux_airgap() are saved in the Magnetics.run(). 

The idea is to return all the outputs as dict (out_dict), for example out_dict["Br"] contains Br.

Then we implement a Magnetics.store_output(out_dict) method that will store all standard outputs in OutMag if they are in the dict.

Finally, MagFEMM and MagElmer will have their own store_output() method, which will first call Magnetics.store_output(out_dict) and then store the additional data. 

That way it becomes obvious where output storage should be implemented, and there is no problem in the future to add other outputs to the MagFEMM/MagElmer model.

BonneelP:
out_dict contains the matrices (with proper periodicity), store_out creates the corresponding Data objects and set them in the Output object.

SebGue:
Hello @EmileDvs , hello @BonneelP ,

I have nearly finished the a first working version of EEC_SCIM and would like to already include this PR. So could you tell me if there are still fundamental changes to come? Further I will introduce mag.Phi_wind_list corresponding to get_lam_list to store the fluxlinkage of multiple Lamination (2 for now) if you don't mind.

BR, Sebastian

EmileDvs:
Hello @SebGue,

It should be done anytime soon, we are currently running the tests. 

Concerning the idea of having a list of Phi_wind, currents, voltage per lamination I think it is something @BonneelP has been willing to do for a long time. Maybe we should open an issue on this, which can be dealt in the same PR than the other issue (https://github.com/Eomys/pyleecan/issues/211) related to modifications on lamination currents.

Best regards,
Emile

BonneelP:

PR #226, opened 2020-11-07
[BC] number of phases was missing in the n2dq transformation

ajpina:
There was a bug when the machine has more than 3 phases in the n2dq transformation, an error because the matrices shape did not match. Only the parameter 'n' was missing in the function call.


SebGue:

PR #228, opened 2020-11-07
[NF] get_lam methods

SebGue:
This PR is related to #212 and add some convienence get_lam_... methods


BonneelP:
Hello Sebastian,

Thank you for this PR :) Should we merge it now or do you plan to extend it ?

Best regards,
Pierre

SebGue:
Hello Pierre,

you should merge it, although the Lamination method get_lam_label is not included yet.
The reason is that this method would utilize the 'parent' property, which may result in some issues (in really rare cases I have to admit).
Therefore I will also open an issue. In short: I think we have to revise the 'parent' property and it's usage since pyleecan object instantces can be set at different places which may result in unpredictable issues, e.g. 

```
stator0 = LamSlotWind()
stator1 = LamSlotWind()
machine1 = MachineSIPMSM(stator=stator1)
machine2 = MachineUD(lam_list=[x, stator_0, stator_1, z]
stator_1.get_lam_label() # will return 'Stator_1' which is not true regarding machine1
```

BR Sebastian

BonneelP:
Hello Sebastian,

Thank you for your PR. Normally we also set the parent for list of pyleecan type, so parent should work for MachineSIPMSM and MachineUD. The issue is when you want to reuse a object somewhere else. For instance when we reuse the InputCurrent object to define the operating point in InFlux. 

Best regards,
Pierre

PR #230, opened 2020-11-09
[CO] Move store method from Magnetic models to OutMag object

EmileDvs:
This PR further improves how Magnetics outputs are saved in OutMag. The previous methods "store_output" of Magnetics models have been merged in a new method "store" of OutMag object.

That way, all the variables are saved the same script, the same way for any model and without redundancy, provided that they have been stored in the out_dict().

The Time_Tem axis has also been added to the comp_axes of Magnetics since torque is a Magnetics standard output.


BonneelP:

PR #232, opened 2020-11-11
[WP] Added boundary condition lines to Gmsh file

ajpina:
This is work in progress to add boundary condition lines into Gmsh file aiming to be used for Elmer Coupling.


SebGue:
Hello Alejandro,

that already looks really good. Although this is still work in progress, I will merge it, so I can contribute to it (without pushing to your master).

Best regards,
Sebastian

PR #233, opened 2020-11-11
[CO] add label "Left" / "Right" to boundary lines for easier identification

SebGue:


BonneelP:
The comment in pyleecan/Functions/FEMM/__init__.py is a good idea, the code is more clear now!

PR #240, opened 2020-11-17
add method gen_elmer_mesh to MagElmer class

SebGue:


SebGue:
Since this doesn't regard 'productiv' code, I merged this PR without review as an exception.

Issue #243, opened 2020-11-18
[CO] Improve Output standards for better storing and saving


EmileDvs:
Hello,

Currently in pyleecan, for a specific physical module, all outputs that comes from any model are stored in the corresponding Output object. We have recently introduced a new standard in the Output class: the store() method wich enables to store standard outputs coming from a model inside the Output. For example, Magnetics.run() stores magnetic outputs using by calling OutMag.store(out_dict), where out_dict is a dictionnary containing all the outputs to save. 

Since we are developing a new magnetic model (MagElmer) and we will probably develop others, the fact that we store the outputs from all models in the same OutMag can become messy, because the OutMag will indeed contain the standards output (Flux density, Torque etc.) but it will also contain outputs that are only relevant to the model (ex: FEMM_dict for MagFEMM, that has been renamed "FEA_dict" on purpose to stay as much general as possible). 

We propose to further improve the way outputs are stored by creating **a new abstract class OutMagInternal** that we set as "internal" attribute in the OutMag object. Then we can create **child objects of OutMagInternal** which will contain the useful outputs related to a specific model (for example: OutMagFEMM for MagFEMM that will contain the FEMM_dict).

Then, we also want to improve the way results are saved. Currently in pyleecan, we can save results in two ways: 

- save all the outputs in the output_list or none of them (VarSimu.is_keep_all_output) 
- save needed data using datakeepers (VarSimu.datakeeper_list) 

We propose to introduce the complementary tool to datakeepers to clean data that is not needed. To do so, we introduce the **"clean_level"** attribute in the VarSimu object and a new standard method **"clean()"** to the Output class, so that the clean_level value determines which output to keep or clean in the Output object of each physics before saving the simulation Output. 

Please let me know if you have any remarks or question about this issue. I am going to start a pull request as soon as possible so you can progressively see the new improvements.

Best regards,
Emile

PR #244, opened 2020-11-18
[CO] Improve Output standards for better storing and saving

EmileDvs:
This PR aims at improving output standards for better storing and saving as developed in issue #243.


EmileDvs:
This PR is work in progress

PR #246, opened 2020-11-20
MagnetType15 - polar bottom, curved top, parallel sides

SebGue:
Hello Pierre,

although you are preparing the unified slots, I would like to add an other magnet type.
The implementation is not perfect, because the actual slot is not parallel. Nevertheless it may be useful to some.

BR, Sebastian

### DIFF
--- a/Tests/Methods/Machine/test_Magnet_Type_15_meth.py
+++ b/Tests/Methods/Machine/test_Magnet_Type_15_meth.py
@@ -1,0 +1,145 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from pyleecan.Classes.LamSlotMag import LamSlotMag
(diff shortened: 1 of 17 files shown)


BonneelP:
Hello Sebastian,

No problem for that, in fact that way we can challenge the organization to make sure that this magnet can also be defined (which should be the case). Can you add some test so I can be sure that the new organization won't break your magnet ? 

Best regards,
Pierre

BonneelP:
Hello Sebastian,

I see that there is a commit "MagnetType15 top curve bug fixed", should I merge this PR now ?

Best regards,
Pierre

SebGue:
Hello Pierre,

no these are other CC I added meanwhile. I still work on the unit tests. But with my next commit you can merge the PR.

BR S

BonneelP:

PR #258, opened 2020-12-03
[CC] get ElmerGrid complete path for subprocess.Popen call

SebGue:
+ black


BonneelP:

PR #264, opened 2020-12-09
[WP] Creates elmer settings files for magnetic analysis. 

ajpina:
Additional entries are added to write the settings file used by ElmerSolve to run magnetodynamic2D solver.


BonneelP:
Hello Alejandro,

Thank you for your contribution. Your PR looks fine to me but sadly I can't merge it because of a conflict on gen_elmer_mesh. Can you solve it ?

Best regards,
Pierre

SebGue:
Hello Pierre, I guess it is due to some of my last PRs, so I will care about that. BR Sebastian

SebGue:
Hello Alejandro, it seems that there is something wrong with your last commit. There are 2000+ empty files. 
I would suggest you to create a new branch to store your current work, go back to master, reset it to the commit before and do the PR again (i.e. without this last commit). If you want to I can also give you the git commands to do so.
After that we can care about your last commit. Maybe there is only some 'empty' difference due to line endings.
Best, Sebastian

SebGue:
Just one more question. Did you use black? Maybe that could also solve the problem (with one more commit).

PR #265, opened 2020-12-10
[NF] Magnetic Force Module

Superomeg4:
This pull-request makes a review of the magnetic force calculation in Pyleecan. 

1) [CC] The notion of Air-Gap Surface Force (AGSF) is introduced: AGSF is the results obtained when applying the Maxwell stress tensor in the middle of the air-gap. The corresponding variables "out.force.P" have been renamed as "output.force.AGSF". There are probably some tests where I have forgotten to renamed variables. Please notice me if this is the case.

2) [NF] The computation radius for the outputs of the MagFEMM module, and so for AGSF, can be enforced. 

3) [NF] The AGSF transfer algorithm has been implemented. It allows to re-compute the AGSF on a different radius without further simulation based on analytical equations from the publication https://www.degruyter.com/view/journals/phys/18/1/article-p658.xml?tab_body=abstract. 

4) New tests have been implemented to check FFT2 accuracy for magnetic field and AGSF.


SebGue:
Hello Raphaël, that's great! I also read some 'old' publication of you on that topic, but had no time to ask some questions especially on VWP :-) Is this method also included? I would like to check the forces on the rotor and if they have some relevancy compared to centrifugal force. BR Sebastian

Superomeg4:
Hello @SebGue,

The VWP is not included. If you want to compare the order of magnitude of magnetic forces vs. centrifugal forces, it should be easy to do it with the AGSF, it is just an issue of orientation. If you want to open an issue about it, I could implement this part.

Superomeg4:
I am probably going to add a Jupyter Notebook in the tutorial section to explain this PR.

PR #266, opened 2020-12-10
[BC] Update requirements.txt

Superomeg4:
There is a bug in numpy 1.19.4 -> Temporary downgrade requirement.


SebGue:
@Superomeg4  Could you tell what the actual bug is.

Superomeg4:
I can't import PySide2:
** On entry to DGEBAL parameter number  3 had an illegal value

but apparently there are other librairies which have an incompatibility with the last release of numpy:
https://github.com/twintproject/twint/issues/1030

PR #270, opened 2020-12-11
[NF] EEC_SCIM - Electrical Simulation of SCIM

SebGue:
I will just PR this now and come back for the details later. I hope there are not too much bugs in it :-)
@BonneelP Maybe someone of you could review this since there are also some changes to actual code, e.g. 'Phi_wind' prop. on OutMag.


BonneelP:
Hello Sebastian,

This PR looks great :) Thank you for this contribution :)
Just one remark: can we use a dict for Phi_wind instead of a list ? The key would be "Stator" or "Rotor_1" for instance. That way the dict would contains only the key that were computed and the Stator data can always be access with Phi_wind["Stator"].
What do you think of that ? 

Best regards,
Pierre

SebGue:
:-) Thank you. I also that idea but forgot meanwhile with all that details, branches, ...

SebGue:
Hello Pierre, 
I think I'm done for now.
BR Sebastian

BonneelP:
Hello Sebastian,

The PR is Ok for me. Should we merge it ? 

Best regards,
Pierre

SebGue:
@BonneelP Yes you can! :-)

BonneelP:

Issue #231, opened 2020-11-09
[NF] Linear elasticity simulation (coupling with Elmer)


SebGue:
Hello,

I would like to announce that I will try to do a first linear elasticity simulation implement with coupling to Elmer.
A detailed discussion will follow as I (hopefully) proceed with this feature within the next weeks.

BR, S.

SebGue:
Hello,

meanwhile I made some progress on this issue and wanted to show my first promising results. Actually the Elmer case files are setup manually. Next I will setup respective pyleecan class and automate the the complete simulation process.
This is also the point where some help is appreciated :-)
- Discuss the implementation in general and in detail (when I'm done with it)
- Get the results (e.g. vtu file) into pyleecan for post processing (@Superomeg4 Maybe you could help here?)
- Discuss about the pyhsics and model assumptions (e.g. senseful boundary conditions)
BTW: At the moment I am mainly interessted in centrifugal force calculation as you can see from my results :-) but magnetic force calculations may be derived from that easily.

Fig. 1 - Disc rotating at 10 krpm (analytical verification can be found [here](http://www.elmerfem.org/forum/viewtopic.php?p=13780#p13780))
![Disc_Centrifugal_10000rpm](https://user-images.githubusercontent.com/43440731/100363089-8a179f00-2ffc-11eb-9174-75918e7666cc.png)

Fig. 2 - slightly modified "IPMSM_A" rotor @ 10 krpm - vonMises stress - Elmer
![IPMSM_PerBC_fixed_N_disp_vonMises_cont](https://user-images.githubusercontent.com/43440731/100363777-7e78a800-2ffd-11eb-97e1-f14840c24113.png)

Fig. 3 - Ref. calc. - vonMises stress - Fusion360 
![Fusion_IPMSM_A_ohne_Kleber_gleitend_vMises](https://user-images.githubusercontent.com/43440731/100364425-67868580-2ffe-11eb-9f6a-d8ef68335761.png)

Fig. 4 -  slightly modified "IPMSM_A" rotor @ 10 krpm - displacement - Elmer
![IPMSM_PerBC_fixed_N_disp_Displacement](https://user-images.githubusercontent.com/43440731/100363915-b1bb3700-2ffd-11eb-8344-1bfa84f3545e.png)

Fig. 5 - Ref. calc. - displacement - Fusion360
[...]

PR #271, opened 2020-12-12
[WP] Added the call to ElmerSolver.

ajpina:
Hi,
This is the same PR as before, after I spent sometime fixing my messed up github fork. Now, I should be back on track.
Best,
Alejandro


Superomeg4:
Hi Alejandro ! 
Great job. How much additional work would it be to get the full FE results (B,H, \mu in every cells) ? 
I can help you building the MeshSolution object, such as in MagFEMM. 

BR 

Raphaël

SebGue:
Hello Raphaël,

I started a convenience class in Public\StructElmer to do the IO with Elmer. In get_meshsolution there is already some code (that has to be moved to the Elmer class) to get the results (see also [here](https://github.com/Eomys/pyleecan/issues/231#issuecomment-743017653)). 
I think this would be the best starting point, if you got some time to spend :-)

BR Sebastian

SebGue:
Thank you for your contribution :-)

PR #274, opened 2020-12-14
[BC] hotfix to load/save OutMagFEMM

SebGue:


BonneelP:
Thank you for fixing that :) 
We still need to solve the main issue with parallelization, I will work on it this week normally.

PR #277, opened 2020-12-16
[NF] Compute Demagnetization

SebGue:
This PR add the get_demag method to OutMag class to get the magnets surface area that exceed a given field strength.


SebGue:
Hello @Superomeg4, could you have a look. I introduced some functions on MeshSolution that could may also be methods but I am not quiet sure. BR Sebastian

pyleecan/Methods/Mesh/MeshMat/get_cell_area.py
```diff
@@ -1,22 +1,39 @@
 # -*- coding: utf-8 -*-
+from numpy import abs, newaxis
 
 
 def get_cell_area(self, indices=None):
-    """Return the area of the cells on the outer surface.
-
+    """
```
Superomeg4, line 6: This method should be defined in the corresponding RefCell (i.e. Triangle3).
Moreover, this type of operation can be performed using jacobian product (so it could be a method of RefCell class).

pyleecan/Functions/MeshSolution/get_indices_limited.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+
```
Superomeg4, line 2: I am not sure why you need this method. Is it possible to directly use get_group method ?

SebGue: It's because it only return indices of cells thats field values are within a given range.

pyleecan/Methods/Mesh/MeshMat/get_cell_area.py
SebGue: I will try to fix that.

pyleecan/Functions/MeshSolution/get_indices_limited.py
Superomeg4: OK. It would be interesting to integrate this feature in get_group latter. ;-)

pyleecan/Methods/Mesh/MeshMat/get_cell_area.py
Superomeg4: Don't worry about the jacobian product, I will do it latter. Just put the ShoeLace algorithm in the Triangle3 class, and call self.cell["Triangle3"].get_cell_area

SebGue: :-) Perfect. Thank you!

SebGue:
@Superomeg4 Could you have a look at it again?
On my side I think I'm done.

Superomeg4:
OK for me :-)

BonneelP:

PR #279, opened 2020-12-16
Rewriting bug messages slot

BenjaminGabet:
Error messages are now correct and tell what's wrong


BonneelP:

Issue #169, opened 2020-09-29
[NF] Unification of SlotMag and SlotWind


BonneelP:
Hello all,

As mention in issue #158 we need to find a way to unify SlotMag and SlotWind. Otherwise we will have duplicate slot for magnet and winding (rectangular slot with winding or magnet for instance). 

This would require to rethink how both slot types are handled and redefined/rename all methods:
- The Slot would now have a "special area" (looking for a proper name, pocket ? filled zone ? **active area** ?)
- All method should be renamed accordingly and should be unified according to the current SlotWind methods
- SlotMag methods need to be simplified by having the "special area" dimensions as property of the slot object instead of the magnet (as we already do for HoleMag)
- The method of the abstract SlotWind class would be moved to the Slot class (then all slot need to have a "special area" that can be the full slot)
- Instead of build_geometry_wind a method "get_area_special" would return the corresponding surface object. Then the split method of issue #24 would enable to split the surface in the corresponding winding layer (or even magnet layers, why not).
- For now winding is a property of Lamination and magnet of the Slot. Both should be in the corresponding Lamination daughters. 
- The possibility to have several magnet to define a same pole should be compatible with this new organization by moving "W3" to LamSlotMag and still using a list of magnet. Maybe we can also have a LamSlotMultiMag class for that ? 
- The methods "has_winding" and "has_magnet" would be useful to add. Maybe we can use them for check in the machine instead of enforcing Lamination types in the machine classes. 

On a side note, one day we should also consider the possibility to use either Slot or Hole regardless of the machine type (induction machine with closed slot, or spoke PMSM). We won't consider this for now, but having winding and magnet in Lamination would be a first step (although HoleMag objects have magnet as properties). 

What do you think of this proposal ?
Best regards,
Pierre

BonneelP:
SlotWind classes are now easier to define thanks to the possibility to cut surfaces (by the way, this cutting method needs to be improved to return both "top" and "bottom" part in one call). 

Here is a proposal for the renaming of the slot methods:
- SlotWind.comp_height_wind => Slot.comp_height_active
- SlotWind.comp_surface_wind => Slot.comp_surface_active
- SlotWind.comp_radius_mid_wind => Slot.comp_radius_mid_active
- SlotWind.comp_angle_wind_eq and SlotMag.comp_angle_opening_magnet => Slot.comp_angle_active_eq
- SlotWind.get_surface_wind => Slot.get_surface_active
[...]

PR #283, opened 2020-12-23
[CO] Merging SlotMag and SlotWind

BonneelP:
Hello all, 

I have finished my work on the SlotMerge branch :) As a reminder the point was to merge SlotMag and SlotWind into Slot to answer #169. 
All the magnets have been converted to SlotMX classes. We managed to create a SlotUD2 class that directly take the active surface as a property and produce the following FEMM simulation: 
image
The GUI and the Tests also have been updated. Now the GUI Magnet step is mostly the same thing as the usual slot one (except for magnet material and magnetization).
LamSlotMag have been reorganized to have only one magnet instead of a list of magnet but one can easily create a LamSlotMagMulti class to get the previous behaviour. To go further one can also investigate splitting magnets the same way we do for winding. 
This PR also solve the import of some dxf Arc and introduce a new "BoreUD" class (that one day will go along a dedicated dxf import widget).

The schematics in the GUI have not been updated (W0 is missing on all SlotMX) but we plan to work "automatized" schematics in a near future. Finally there is still some docstrings and comments that need to be updated to rename "winding" into "active"

I won't merge this PR before a while to let everyone inspect/improve it. Note that it break retro-compatibility: all machine using LamSlotMag object need to be updated (Hmag and Wmag from magnet to slot, update slot type, update magnet type, remove W3 from slot, move magnet from slot to lamination). All existing validation machine have been updated. Once this PR merged, we plan to do a new release so that the new standards are used as soon as possible. This release will be 1.1.0 and the roadmap will be updated afterwards. 

Best regards and merry christmas,
Pierre


SebGue:
Hello Pierre, 
thank you for that PR. I guess this was a lot of work. I hope I can finally have a look at it the next days. :-)
Is there some tutorial on how to split windings? I think I missed that.
BR and Merry Christmas, Sebastian

BonneelP:
Hello Sebastian,

The "active surface" is automatically split in the file pyleecan/Methods/Slot/Slot/build_geometry_active.py and some improved versions can also be available in the SlotWX classes (for instance pyleecan/Methods/Slot/SlotW10/build_geometry_active.py).
As a side note: the method to split can be improved by changing the return of the split_line method to always return both top and bottom part (to avoid calling twice the method). I will open an issue or work on that once I'm back from holidays.

Best regards,
Pierre

BonneelP:
Hello all,

I have added a new experimentation in this PR. The "SlotW10.plot_schematics" method to plot the schematics either on default or current slot. Can someone check if the general file organization is Ok or if there is any other feature or variable to add ? If this template is Ok I will start to add this method for all the other slots (which can take a while). Here is how it looks like with default slot and a save_path:
test

Best regards,
Pierre

BonneelP:
Hello, 
As requested, I have added a function to convert the init_dict when loading an old version of a SIPMSM machine. This function (and the corresponding test) will be extended after each new reorganization. 

Best regards,
Pierre

PR #287, opened 2021-01-06
[CC] demagnetization

SebGue:
This PR should be merged after IronLoss PR to correct group name.


BonneelP:

PR #291, opened 2021-01-08
[NF] ElmerResultsVTU helper class to load results

SebGue:
This PR add a new Elmer class to load the vtu results into pyleecan. 
Since this class has only one method, it could have been done with a function also. But because there are also other convenient Elmer classes, I decided for a class over a function.

### DIFF
--- a/pyleecan/Methods/Elmer/ElmerResultsVTU/get_meshsolution.py
+++ b/pyleecan/Methods/Elmer/ElmerResultsVTU/get_meshsolution.py
@@ -1,0 +1,124 @@
+# -*- coding: utf-8 -*-
+from numpy import arange
+from meshio import read
+from os.path import join, split, splitext
+
+from SciDataTool import DataTime, Data1D, VectorField
+
+from ....Classes.SolutionData import SolutionData
+from ....Classes.SolutionVector import SolutionVector
+from ....Classes.MeshSolution import MeshSolution
+from ....Classes.MeshVTK import MeshVTK
+
+from ....Methods.Elmer.ElmerResultsVTU import ElmerResultsVTUError
+
+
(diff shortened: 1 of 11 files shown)


ajpina:
Hi @SebGue ,
What about Cell Data? I see that get_meshsolution() in this PR is loading only Point Data.
Best,
Alejandro

SebGue:
Hello Alejandro,

you are right. Actually this is not implemented. Do you have such data file?
If you upload it, I try to add this feature as soon as possible.

Best regards, Sebastian

ajpina:
Please check out my initial attempt based on your code PR #292 `get_meshsolution` method. It's working so far with my test case but not sure whether is general enough yet.
Best,
Alejandro

pyleecan/Methods/Elmer/ElmerResultsVTU/get_meshsolution.py
```diff
@@ -0,0 +1,124 @@
...
+from ....Classes.MeshSolution import MeshSolution
+from ....Classes.MeshVTK import MeshVTK
+
+from ....Methods.Elmer.ElmerResultsVTU import ElmerResultsVTUError
+
+
+# TODO add groups, see get_meshsolution of MagFEMM
+
+
+def get_meshsolution(self):
```
Superomeg4, line 19: To be more coherent with MagFEMM, I would rename as "build_meshsolution" instead.

SebGue:
@BonneelP I think you can merge this PR now.
@ajpina Sorry for my late reponse. I had a look at your PR. I looks good. As soon as I got some time, I will merge your get_meshsolution to the general ElmerVTU class.

BonneelP:

PR #292, opened 2021-01-10
[WP] Added functionality to MagElmer Class.

ajpina:
This PR adds some surface permanent magnet geometries to the geometries that can be solved so far via Elmer coupling, enhancing pyleecan capabilities to run on Linux and Mac OS platforms.


BonneelP:
Hello Alejandro,

Thank you for this PR :) I finally have some time to investigate it and here are my remarks:
Tests/Functions/test_gmsh2.py: I will soon merge the PR #283, MagnetType10 will no longer exist after that. 
pyleecan/Functions/GMSH/get_sliding_band.py: I'm wondering why you needed to add an exception for MachineSIPMSM ? Normally the definition for the mechanical airgap radius take into account that there can be magnet in the "magnetic airgap" (out.Rint - in.Rext). Can you show us an example of what the previous version of the code wasn't able to handle ?  
pyleecan/Methods/Simulation/MagElmer/solve_FEA.py line 327 & 426 : I have spotted the following lines:

>  conductivity_m = 0.0*1.0 / rho_m

I just report it to make sure that it is normal ;)

Best regards,
Pierre

ajpina:
Hi Pierre,
1) I will check what to do regarding MagnetType10 in my next PR.
2) get_sliding_band.py searches for rotor and stator lines in order to be used for the air gap surfaces. The issue is that air gap faces in SPM geometries will need to include magnet lines as well as rotor lines, that's why it's been used both mechanical and magnetic air gap radius to be able to find the lines. The other problem is the overlapping lines between magnets and rotor core in SPM (Arc magnets) which is still an open issue. The previous version of the code did not subtract the magnet surfaces from the air gap surface when meshing, so it could not be solved by Elmer. This version is WP to solve both IPM and now SPM geometries too.
3) Conductivity is hard-coded to zero for now to remove eddy currents from the solutions since conductors are modeled via coil faces and current density enforced on them rather that actual conductors cross section. I noticed that you have a parameter call `is_eddies` when creating the windings and wondering if it is for the same reason.
Best,
Alejandro

ajpina:
Hi Pierre,
Issue with MagnetType10 has been addressed. 
Best,
Alejandro

BonneelP:
Hello,
Sorry for the long time to answer. I checked your PR and it is Ok for me to merge it. Should I do it now or do you have something else to add ? 

2: Would it help if we create a new line label convention so you can find easily the proper lines ? What we can also do is introduce a get_bore_line_active method (or something like that) that would return the bore lines including magnets (like build_geometry do but without magnet) ? This new method shouldn't be that hard to create with the new architecture (by calling/combining both build_geometry and get_surface_active) and it may no be required for all slots .

3: Indeed in FEMM we use "is_eddies" to change the material definitions cf Functions.FEMM.create_FEMM_circuit_material for instance.

Thank you again for your contribution,
Best regards,
Pierre

ajpina:
Hi,
I am fine with the merge now, will work later on troubleshooting outer
rotor geometries and winding with more than one layer.
Best,
Alejandro

BonneelP:

Issue #301, opened 2021-01-23
[BC] FEMM periodicity


Superomeg4:
Hello everyone,

I have found a bug related to FEMM periodicity thanks to the test "test_FEMM_periodicity". It is particularly obvious on the tangential component of the AGSF (see image below), but it is also existing on the tangential component of the magnetic flux density. It looks like the issue appears when no periodicity is considered. I am going to investigate this and put my findings for the record, so people working on new Magnetic models might avoid similar issues. 

![image](https://user-images.githubusercontent.com/23015350/105574441-dc5b0f00-5d64-11eb-8ddf-e0daa34b78ba.png)

I could think of several possibilities:
- The last angle step value is wrong without periodicity.
- Same but with time steps. 
- It could be because the discretization is too coarse, and/or the sliding band introducing numerical errors. 
- The rotor rotation angle could lack of accuracy when considering the full model.

If you have other ideas in mind, feel free to propose them. 

Best regards,

Raphaël

Superomeg4:
Update: 

The test itself is now working, but there is still a strange behaviour when running test_FEMM_periodicity_angle in test_FEMM_periodicity.py.

If I set Nt_tot = 2*9 in InputCurrent, everything work correctly except the torque:
![FEMM_periodicity_angle_Tem_time](https://user-images.githubusercontent.com/23015350/106495436-f78ff200-64bb-11eb-9721-f1a4f9f6e6f3.png)

The problem disappears if I set Nt_tot = 4*9 :
![FEMM_periodicity_angle_Tem_time](https://user-images.githubusercontent.com/23015350/106495461-fe1e6980-64bb-11eb-8070-a5072567820d.png)

PR #310, opened 2021-02-01
[BC] Periodicity

Superomeg4:
This PR does not make any modification inside pyleecan. 
There is only a correction of test_FEMM_periodicity to reduce calculation time and "make it work". 

Related to #301 .


BonneelP:

PR #312, opened 2021-02-03
[CO] Retro comp for all pyleecan objects that include machines

SebGue:
This PR also include the 'Ntcoil' commit of slot branch.


BonneelP:

PR #313, opened 2021-02-03
[NF] OptiObjective class 

SebGue:
This add the new OptiObjective class that is a pure 'copy' of DataKeeper. It is intended to distinguish optimization objectives from normal datakeepers, e.g. for post processing.
This PR also include some CO to remove duplicate code and add some error handling.


BonneelP:

PR #314, opened 2021-02-03
[BC] Q=1/2 winding + reverse winding

SebGue:
This PR corrects the phase order of Q=1/2 windings and corrects the reversal of tooth windings. It also trys to increase readability by removing some line breaks.


BonneelP:

Issue #22, opened 2019-04-18
Improve assign_FEMM_surface for HoleMagnet


BonneelP:
There are two things to improve in assign_FEMM_surface regarding the HoleMagnet:
- For now the magnet are always set according to hole[0] but in the case of several holes it would require to read the label to select hole[ii]. The convention could be that in build_geometry of LamHole we update the hole label to have RX with X the position of the hole in the hole list of the lamination. 
- We need to add a method comp_angle_magnetization to compute the angle of the magnet for parallel magnetization. It would take as argument a bool is_north and return a tuple corresponding to the magnetization of each magnet (None if the magnet is not there).

jnddong:
For the second point: is it necessary to add an option to define your own magnetization angle?

BonneelP:
Hello,

Regarding the second point the idea is to find a convenient way to get the correct angle for parallel magnetization because it depend on the Hole orientation/geometry. For instance for Hole 50:
![image](https://user-images.githubusercontent.com/20533034/58253429-bcda3a80-7d68-11e9-8cf3-820669964981.png)
it would be the angle to have the magnetization like the H2 arrows. We already compute that in Functions/FEMM/assign_FEMM_surface (line 63 to 92). The idea is to provide a method in the HoleMX classes to get directly the correct angles to replace the:
`if type(lam.hole[0]) == HoleM50
....
if type(lam.hole[0]) == HoleM51
....`
by something like:
`angle_list = lam.hole[0].comp_angle_magnetization()`
Which is more coherent with the object oriented approach and easier to adapt when len(lam.hole)>1.

So for parallel magnetization there is no need to add a new parameter (otherwise the user will have to define this value to set the correct magnetization). However, once we have the comp_angle_magnetization method, it could be interesting to add a parameter in the Magnet object to enforce the returned values. That way it would be possible to enforce easily the magnetization of any magnet in any direction for research purpose. 

Best regards,
Pierre

BonneelP:
As a side note, I'm only referring to parallel magnetization. So maybe the method name should be comp_angle_parallel_mag(is_north=True). 
For now we only have parallel and polar magnetization but there will be some other type and it would also be interesting to adapt assign_FEMM_surface to avoid the 
```
if "Parallel" in label
...
if "Radial" in label
...
```

BonneelP:
Hello all,

[...]

PR #323, opened 2021-02-19
[NF] comp_magnetization_dict

BonneelP:
This PR solve the issue #22. Now each hole have a comp_magnetization_dict method and Hole class have a magnetization_dict_enforced property that can be used for HoleUD. I have added a test where I was able to run in FEMM the prius with HoleM50 and with the convert_to_UD equivalent. 

I won't merge this PR for now, I want to carry on the work to be able to define magnetization_dict with the Hole_DXF GUI. Meanwhile, feedback welcomed on the magnetization_dict :)


BonneelP:
Hello all,

I'm done with this PR. So now in the Hole_DXF GUI one can select the magnetization direction of each magnets. I have added the BMW_i3 machine that was created with the Hole_DXF gui. For both DXF GUI, I have added "Reset view", "Cancel Selection" buttons, a scaling factor (for dxf in [mm] for instance) and some Doc. We should produce some video tutorial about these two features in the near future. @SebGue do you think you will have the time to check this PR ? 

Best regards,
Pierre

SebGue:
Hello Pierre, I will take the time and go through it after lunch. BR S

pyleecan/Methods/Geometry/Line/comp_normal.py
```diff
@@ -0,0 +1,39 @@
...
+    Normal point is "on top" (begin=O and end on Ox)
+
+    Parameters
+    ----------
+    self : Line
+        a Line object
+    return_type: int
+        0: (Zbegin+Zend)/2 (even for arc) and the normal point as tuple
+        1: Z2 - Z1 (with Z1 and Z2 the points from return_type 0)
+        2: the angle of the vector of return_type 1 according to Ox [rad]
```
SebGue, line 15: The varialbe return type seems strange to me. Since there are essentically only 3 relevant quantities (Z1, Z2, ang), I would return them all. Then every on how needs type 1 can do Z2 - Z1 themselve. 
Futher one can also 'filter' the return, e.g. comp_normal[2] for the angle only.

pyleecan/Methods/Slot/Hole/comp_magnetization_dict.py
```diff
@@ -0,0 +1,36 @@
...
+    if self.magnetization_dict_enforced is not None:
+        if return_type != 2:
+            raise Exception(
+                "For Hole, only return_type=2 (float) is available for comp_magnetization_dict"
+            )
+        mag_dict = self.magnetization_dict_enforced.copy()
+        if not is_north:
+            for key in mag_dict.keys():
+                mag_dict[key] += pi
+        return mag_dict
```
SebGue, line 34: This seems inconsistent with HoleMxx, since there is modification of the enforced values, although this is the abstracts class method only.

pyleecan/Methods/Slot/HoleM50/comp_magnetization_dict.py
```diff
@@ -0,0 +1,44 @@
...
+    Parameters
+    ----------
+    self : HoleM50
+        a HoleM50 object
+    is_north: True
+        True: comp north magnetization, else add pi [rad]
+    return_type : int
+        0: Normal as tuple of complex
+        1: Z2 - Z1 (with Z1 and Z2 the points from return_type 0)
+        2: the angle of the vector of return_type 1 according to Ox [rad]
```
SebGue, line 20: See 'comp_normal'.
If you keep the current implementation with retrun_type, the return_type isn't checked if enforced dict is not None (despite in Hole comp_mag. method).

SebGue (commented):
Haven't tested the functionallity, just had a look at some of the code.

pyleecan/Methods/Geometry/Line/comp_normal.py
BonneelP: Hello,

I agree, return type was more an experimentation at the beginning. But now I think that we can just change it to return always the second option because we can't have option 0 or 1 for HoleUD.

pyleecan/Methods/Slot/Hole/comp_magnetization_dict.py
BonneelP: Indeed. Do you think that we should say that magnetization_dict_enforced enforce the "north direction" and apply this modification in all HoleMx ? The other solution is maybe to remove this option (I think it is unused anyway). What do you prefer ?

pyleecan/Methods/Slot/HoleM50/comp_magnetization_dict.py
BonneelP: Same, I think that the cleanest solution would be to remove return_type.

pyleecan/Methods/Slot/Hole/comp_magnetization_dict.py
SebGue: I think this shouldn't be removed but disabled for now. 
We should discuss in a dedicated issue, what this feature can do, e.g. also utilize the FEMM magnetization by string feature (see Hallbach). Further it should be avialable for surface magnets also.

BonneelP: I was talking about the "is_north" option ;) Indeed the magnetization_dict_enforced  will be quite helpful in the long run. So I will comment the is_north part in all holes for now.

SebGue: So regarding enforced values, the question is, will it be absolute values or offset to the computed. I think offset would be more accessible, since the user dont have to care about the actual magnets position.

BonneelP: I think that we should either use "absolute definition " either change the property name to "magnetization_dict_offset". 
With "absolute definition", we can create the offset by calling the comp_magnetization_dict and updating this dictionary. Then for HoleUD and all magnets we take directly the value within the dict. 
With the "offset definition", we can say that the "computed part" of HoleUD is 0 and enforce the offset. The offset definition is interesting when considering sensitivity/fault study since "0" would be the reference direction. I agree that we will need to extend this definition to surface magnet and one day we may also want to be able to change the magnetization any of the "Zs" magnets (by using Magnet_R0_T0_S0 notation maybe). For this case again, "offset definition" would be more convenient. So I will update the code to the "offset definition".

BonneelP: Then with the "offset definition", we can use "is_north" to add a "pi offset" ;)

BonneelP:
I think that I have made all the changes that we discussed and the tests are still Ok.

SebGue:
So from my side, you could merge this PR. BR and have a nice WE. S

BonneelP:
Thank you for your help :)

PR #324, opened 2021-02-22
[BC] and [CO]

Superomeg4:
1) Updating workflow which save FEMM mesh and solution data
2) Correcting a minor bug in class generator


BonneelP:

PR #327, opened 2021-02-23
[BC] Compatibility with FEMM parallel

Superomeg4:


BonneelP:

PR #328, opened 2021-02-24
[CO] WGuiOption

SebGue:
This PR allows to use the Options Widget without actual instances of DMachineSetup and MatLib.


BonneelP:

PR #329, opened 2021-02-24
[BC] WPathSelector

SebGue:
Fix bug where canceled selection dialog sets path to empty string.


BonneelP:

PR #330, opened 2021-02-24
[BC] WPathSelector

SebGue:
Fix bug where canceled selection dialog sets empty path.


BonneelP:
Duplicate of #329

SebGue:
@BonneelP This is actually not a duplicate since it is for WPathSelector while #329 is for WPahtSelectorV

BonneelP:
Indeed, sorry I didn't noticed the difference. Thank you :)

PR #338, opened 2021-03-15
[BC] fix HoleM57

SebGue:
There are some label issues with HoleM57 that raised some FEMM model build issues. With this PR this should be fix. 
Nevertheless, I seems we need to discuss the line labels with respect to 'get_mesh_param' FEMM function.


BonneelP:
Thank you for this correction. 
Maybe on this Friday I will try, once and for all, to write the "label guidelines" documentation and update the code accordingly. I will open an issue this week to gather all the previous discussions.

PR #341, opened 2021-03-16
[CO] avoid grid toggle + plot title only if needed

SebGue:
If one wants to plot multiple data (e.g. 2D and 3D) into one figure, this PR fixes grid toggle (on 2D) and unwanted title overriding.


BonneelP:
In a near future, we plan to move these generic plot functions to SciDataTool (we use the same plot in Mosqito and it makes sense to have generic plot in SciDataTool, in particular if we reuse the same syntax). It should only change an import in pyleecan.

PR #348, opened 2021-03-19
[CO] Init_environment

BonneelP:
This PR improves the initialization of pyleecan. Now the logger are properly setup and when loading the config_dict we check its version to know if we need to update the files in the APPDATA (to update matlib for instance). 
Magnet1 was also corrected and the GUI now better handle None shaft and frame.


SebGue:
Hello Pierre,

I just realized (since I had some issues with config_dict), that the MatLib and Machine path is always overriden. Why's that?

Best regards, Sebastian

BonneelP:
Hello Sebastian, 

It makes sense in Manatee where we have two different path: APPDATA for the reference (to be copied to the workspace) and workspace for the one used in the GUI. So now that you mentioned it, we can just enforce these path when they are missing. I will update it.

Do you have any other problems with the config_dict ? 

Best regards,
Pierre

SebGue:
I had issues with one of my internal GUI classes, but I couldn't reproduce it with the public. I fixed it by saving the changes in the paths to the dict.
I.e. I imported config_dict (as in any other method/function) but the paths were empty. Don't know why.

BonneelP:
Maybe it is linked to the fact that we use the same object in all the import. So if the config_dict is imported in a file and modified it will also be modified for all further use.
Can I merge this PR ?

SebGue:
Yes you can :-)

PR #349, opened 2021-03-19
[WiP] save compressed json

SebGue:
This is a WiP PR to directly discuss some changes that are needed to do to the save_json/load_json functions.


SebGue:
@BonneelP Do you mind if I make the class_to_split an input of save_json.

SebGue:
... just forgot. The first 'main' goal is to allow switching between dump (file output) and dumps (string output for further processing) method of json.

BonneelP:
Hello,
Maybe it is just me (since it is a long time since I've worked on these methods), but I think that the save (for both json and h5) are overly complicated. When there is an error it is sometime hard to debug (just knowing which variable raised the error is sometime hard). So I don't mind changing some argument/method as long as we take the opportunity to improve the global structure for better understanding/debug. Maybe it is just a matter of renaming/reorganizing the method. This is a complex matter (this method is very flexible and it is great :)) so it is normal to have script that handle a lot of complexity but if we can simplify the highest function that would be great :) Maybe I can help you on this topic.

Best regards,
Pierre

SebGue:
So I'm glad. I also think it's very complicated. 
Do you think we have to allow dict as input to save_jsons obj input?
I would go for the following structure: 
Serialize the object and collect every (sub) object that needs to be saved seperately in a list. Then we only need to use dump in one place.

BonneelP:
Yes I agree that a structure convert/parse/handle strange types then dump would be more understandable. We can then bypass the convert part if we directly provide a proper dict/serializable object. So Ok to use it to save dict as well. 
When serializing a simulation we can set "machine_name.json" in the simulation dict and add the machine object to a "to_save_separately" list (depending on is_folder) and call recursively the save method. 
Maybe we can take the opportunity to redefine the "is_folder" what about providing a list of type that we want to save separately ? (Machine, Material, and maybe Mesh or VectorField depending on the application).

SebGue:
:-) I think I wasn't too clear on the 'dict thing'. I thought if we only allow pyleecan obj., we could remove some lines from build_data.
But I will leave that open, since it is only a detail.
BTW: I found some but saving multisimulation since the machine and material names will be the same for every split object. So machines, ... will be overriden.

SebGue:
Hello @BonneelP ,
I think I am at a 'critical' point now. I got the list of split object with their requested file names now. So I could reduce the code to use dump function in one place only.
Nevertheless I was wondering if it would make sense to use subfolders (e.g. named by the class type) to store the data. This could be beneficial if you think of saving multisimulation with 'keep_all',i.e. tons of files in one folder. On the other hand it may be too complex regarding implementation and the resulting path trees. 
I could also think of only subfoldering the top level of an object.

Since you seem to have more experiance with save to folder option, I hope you can give me some advice. :-)

Best regards, Sebastian

BonneelP:
Hello Sebastian,

Thank you for your work on this feature, it start to be more understandable. Here are some thoughts to further improve this part of the code (it go further than your PR):
- First a small detail that bothers me: the function order is upside down, the most important function of the file is at the bottom and you have to scroll all the way down to understand what is going one. We need to order the function starting from save_json and then by order of usage/importance. 
- There is a getLogger("Pyleecan"), we need to introduce loggers.MAIN_LOG_NAME and GUI_LOG_NAME=MAIN_LOG_NAME+".GUI", same for simulation logger.
- We have a function "fix_file_name" but in save_json they are further path/name manipulation, this part of the code need to be included in "fix_file_name" (that can be renamed to "setup_save_path" for instance?)
- class_to_split docstring "only if is_folder=True"
- I see that you solved something that bother me: when saving a simulation with is_folder=True we used to have a log entry for each file created. What about customizing the log message ("Saving simulation in folder (5 files)" or something like that ?)
- Regarding tuple I would say that we can serialize them with list. In the init of the corresponding class we should then use the setter to convert list to tuple. 
- I wondering if the split should happen in the build_data function: we are already going object by object and as we have not called as_dict yet we can use isinstance(obj, class_to_split) (class_to_split would then need to be a list of type - maybe need for dynamic import since save is in all classes). Then no need to import class_dict. build_data could then return a list of serialized object to dump. Every time we found a class to split, we (for instance) set "machine.json" in the current obj dict and we call result_list.append(build_data(machine)). It would enable to do something like "for result in result_list: dump". For that loop we "just" need to find a way to handle the save path: either we use "<name>.json"/"<type>.json" (works for machine, material, simulation), either build_data return a list of dict(path and serialized obj). The build_data function won't be that much complicated but it should greatly simplify save_json.
- For me output should be saved in a single .h5 file. But let's try with json. Why are we splitting: to be able to read each interesting part. How should we split: so that the load works easily. I would go with only one example for now: xoutput with is_keep_all_output=True. I would go with the following folder content: output.json (main file with most data except simu and output_list), simulation.json, machine.json, material.json and a folder for each output in output_list (output_X with output, simu, machine, material). We would need to adapt build_data to handle [Output]. Then when loading we need to adapt the setter of output_list to handle string for each element of the list (["output_0","output_1",...]). What is great with that is that each output could be loaded separately and one can read/get each machine variation. We need to know if in output_X/output.json we should put simulation.json or output_X/simulation.json, it will greatly change how to organize the code. 

I hope these ideas will help you, I can further detail/help you implement some of them if needed.
Best regards,
Pierre

SebGue:
Hello Pierre,

I edited your last anwser and added checkboxes. I hope this is okay for you.
I will check the things, that I think I have addressed. But you could also uncheck if you think there is something left.

Best regards, Sebastian

BonneelP:
Hello Sebastian, 

What is the status on this PR, do you need any help ?

Best regards,
Pierre

SebGue:
It's been a long time...
I think I finished 'my points' so far, but there were some points from your list, that wasn't addressed yet.
So if we fix the conflicts and do some tests, we could merge it and maybe make an issue of all open points?!
Best regrads, Sebastian

BonneelP:
Yes sorry, I thought you wasn't finished with your feature :/
Indeed we can validate the current state that add an important feature and come back later for the others points. Anyway we plan to investigate the save/load functions which are taking too long.

SebGue:
You couldn't have known since I haven't told.
Regarding performance (in general not only load/save), do you use profiling?

BonneelP:
We have not worked a lot (yet) with the profiling related packages. We are more working on the algorithmic/scientific side. But for these two functions we have investigated cProfile + snazeviz.

SebGue:
Regarding profiling of load/save, could you show some results, i.e. what are the 'slow' parts of the code.

BonneelP:
Sadly I don't have the report any more. But if I remember well, the "tolist" and init of the ndarray were taking a bit of time. We should further investigate in the near future (in particular by running again the profiler) but I would like to check if we could directly save the ndarray in h5 without the tolist conversion (then it would require a new parameter for as_dict to skip call to tolist).

BonneelP:
Hello Sebastian,

I have merged the latest modifications in your PR and there is an issue with the test "Tests/Functions/test_save_load.py::test_save_load_folder_path". It seems that when saving the Toyota_Prius, "M400-50A" is saved several time although the same object is used several times (it creates M400-50A_00001.json...). I need to investigate a winding bug and then I will try to investigate this test.

Best regards,
Pierre

SebGue:
Hello Pierre,

that was actually a bug of the 'old' save function as far as I know. I.e. if there are different materials with the same name, all that materials were saved to the same file, since the file name is defined by the material name. There were no checks if the materials are the same. (You can check that in the test terminal output where M440 and the magents are written multiple times.)

With this PR there is also no check if objects are the same. They are saved to different files regardless. 
Of course this could be improved, e.g. by processing the 'split_list' to remove existing duplicates of the same object.
Should we include that before merge?

BR, S

BonneelP:
Hello,

Indeed the current version also doesn't include that but I would like to solve this issue before merging (we also had some issue recently on the topic :/). I can work on the topic, I just want to discuss about what would be the best solution. 

Here is some proposal/thoughts on this topic:
- When saving an output/simu/machine in folder mode we need to check that the folder doesn't exist or is empty to avoid renaming problem (simu.machine.name = A, simu.save, simu.machine.name =B, simu.save => folder have both A and B).
- When splitting simu or machine, we know that there will be only one of these object so we can save them directly
- Material can be handled separately by introducing a second "cls_tuple" that create a list of material, then we sort them by name, compare to detect the differences to get new name if required and then the machine object needs to be updated (to save the correct material name)
- We could also introduce a method in machine that would assert that all the material with the same name are identical or do the proper renaming (we already have Machine.get_material_dict). We would then "just" need to call this method before saving the machine and no extra work would be require on save (we can overwrite the json if needed).
- We can also directly remove the split for the Material and split only simulation and machine. The point of spiting is to ease the direct use of the json file (instead of the GUI). A machine with split material is way easier to read (in particular if the material contains its B(H) curve) but when sending a machine file (by email for instance), you must send all the materials as well otherwise you won't be able to open the machine (I can't remember how many time we had this issue). And sending a machine.json file is a manipulation that should happen more than editing the json when there already is a GUI for that. 

What would be the best solution for you ? 

Best regards,
Pierre

SebGue:
Hello Pierre,

I can't give you a complete answer now, since I got no time for now. 
Have you looked at the PR code? There is a list of objects to split 'split_list'. These objects are saved only at the end of the splitting process. So what do you think of editing this list to remove duplicates?

Best regards, Sebastian

BonneelP:
Hello, 

I will have a look today on this topic, I will check this list thank you for the advice.

Best regards,
Pierre

BonneelP:
I saw that the docstring for the parameter "compression" of save_json is missing. It seems that it should be a string with "gzip" to work. Do you plan to add other format or can we rename it to "is_compression" as a bool ? Maybe we can also have "type_compression" (0: nothing, 1: gzip etc).

SebGue:
I would like to have an option for the compression type so "type_compression" is fine.

BonneelP:
Hello,

After working on some tests, I realized that I didn't fully understood how it worked. I assumed that the file name and material name were matching which is not the case and that's great. With that the load process is straight forward, the mismatch problem is solved. There are more files that required if all the materials are identical but I don't think that's an issue since with that the user can specifically interact with the rotor M400-50A for instance.

With this new realization, here are the modification I plan:
- I will add a new parameter "is_delete_old=True" to delete the old json files in folder mode (simu.machine.name = A, simu.save, simu.machine.name =B, simu.save). With log message and Exception. 
- I will investigate how we can change "M400_50A_000001" to "M400-50A_Rotor-0" but I think that it can be done in another PR

I will try to merge this PR today to include it in another release.
Thank you again for this contribution.

Best regards,
Pierre

BonneelP:
Hello Sebastian,

I still have some test to correct but I think that I'm done with my modifications. Do you think that you would have the time to take a look at 77469b4df8d3ba5eb288a33b08e5cc5368f8f7de to check if you agree with the save behaviour ? test_save_load_folder_path highlight the current behaviour:
- Saving in the folder remove the previous files (to avoid machineA, machineB situation)
- When several file have the same name _0000X is added (to keep different version of a same material with the same name)
- If a file to split have a name at None, it is not split (Insulation example). This should only happen for undefined material

On a side note I'm considering adding SlotUD and HoleUD to the split list (as the line_list can be long) but these objects have no name. We will see in another PR.
Thank you again for your contribution.
Best regards,
Pierre

BonneelP:
I have added a new test for save json compressed and all the tests are passing

SebGue:
I will check it.

pyleecan/Functions/Save/save_json.py
```diff
@@ -230,7 +251,12 @@ def split_obj_dict(cls_tupel, obj_dict, folder, split_list, file_ext, logger):
                 cls_tupel, val, folder, split_list, file_ext, logger
             )
 
-        if "__class__" in obj_dict.keys() and obj_dict["__class__"] in cls_tupel:
+        if (
+            "__class__" in obj_dict.keys()
+            and obj_dict["__class__"] in cls_tupel
+            and "name" in obj_dict.keys()
+            and obj_dict["name"] not in ["", None]
```
SebGue, line 258: Hello @BonneelP ,

with these lines you prevent to split unnamed and nameless objects. So it's impossible to e.g. split Input, EEC or Drive objects. 
I guess you told why you did that, but I missed it :-) Maybe there should be another input arg 'save_named_only=False' if you need that restriction.

Best regards, Sebastian

SebGue:
... an additional note: We could also significantly reduce 'normal' json file size by reducing indentation to e.g. 1 or 2 instead of 4.

pyleecan/Functions/Save/save_json.py
BonneelP: Hello,

The purpose was to avoid splitting material with name at "None" (which most of the time means that the material is not defined). I will change the condition to always split the the object without name property. By the way, I will introduce a "name" property in SlotUD and HoleUD to split them as well.
Regarding the json indentation, it is worth the try. I will check what an indentation of 2 looks like.

Best regards,
Pierre

PR #354, opened 2021-03-23
Plot

helene-t:
Remove plot_2D_Data, plot_3D_Data, plot_2D, plot_3D, plot_4D, and use SciDataTool plot methods and functions instead


SebGue:
Could you give me some time before completely removing all plot functions from pyleecan (at least one day)?

helene-t:
Yes of course. You can still use plot_2D, plot_3D and plot_4D if you import them from SciDataTool instead (the syntax is the same), and plot_2D_Data/plot_3D_Data must now be called directly from SciDataTool objects (DataND or VectorField).

SebGue:
If it is not too much affort, I would vote for keep the plot functions the following way. 
The pyleecan plot functions pass the SciData plot functions (e.g. with from SciDataTool import plot_2D, haven't tested if this works) and give some deprication warning that they will be removed with v2 for instance. Otherwise users may be annoyd if they use the plot functions in their scripts and the functions 'suddenly' disappear.
One other minor downside is, that we lose the ability to set colors via config_dict. 
Best regards, Sebastian

helene-t:
Ok, I can leave plot_2D, plot_3D, plot_4D in pyleecan, with a direct call to the equivalent functions from SciDataTool (from SciDataTool.Functions.Plot.plot_2D import plot_2D) and a deprecation warning.

Concerning plot_2D_Data and plot_3D_Data, do you think I should temporarily leave:
- the functions (e.g. plot_2D_Data(out.mag.B, "time") calls out.mag.B.plot_2D_Data("time"))
- or the output methods (e.g. out.plot_2D_Data("mag.B", "time") calls out.mag.B.plot_2D_Data("time"))
- or maybe both?

Concerning the colors, I added dict_2D and dict_3D to automatically pass colors and fonts from the config_dict to the plots.

Best regards,
Helene

SebGue:
Hello Helene,

hard to tell. I would say you should keep everything that could be potentially used by some user. But at least everything that is in the tutorials, as long as they are not updated, should be kept.
Best regards, Sebastian

helene-t:
I already updated the tutorials according to the new syntax. It is included in this PR.

I restored the plot functions, but not the plot methods from the output object because I do not have so much time to work on it. Let me know if you find it really necessary. Otherwise you can merge this PR whenever you are ready (no rush on our side).

SebGue:
Its okay for me to have the functions. So you can merge it.

SebGue:
Just wanted to let you know. I've got some small project where I need to plot some data, so the SciDataTool plot functions are really handy. Thanks for that.

BonneelP:

PR #355, opened 2021-03-24
[NF] getter for ParamExplorer and get_desc for preview

BonneelP:
As we now have the possibility to get the reference in datakeeper, this PR add the possibility to get the reference from paramexplorer (for further plot improvement one day)
Using this new feature, I have added a new method to have a "preview" of VarParam:
image


BonneelP:
I have added some [BC] to this PR, now all the tests are Ok on this branch, I will merge and do a release (for the new plot).
By the way, now we can use DataKeeper of SciDataTool object and save/load the simulation.

PR #366, opened 2021-04-06
Code optimization for Meshsolution module

Superomeg4:
This PR have the following modifications:

1.  Optimized call of get_field method for MeshSolution. It is now looping available transformation on all axes, whatever their type and number. Moreover, there is no more assumption made on the order of the axes for this part of the code. 
2. A little optimization for normals calculation. (It is saving a lot of time when the field is changing but not the mesh). 
3. Automatically select first element of each slices (different from indice and component). It was not working with axis different from time and freqs.
4. Use SciDataTool alike args to select slice of field along given axis. For example:
 `    out.mag.meshsolution.plot_contour(
        "time[2]",
        label="H",
        save_path=join(save_plot_path, simu.name + "_H_time2.png"),
        is_show_fig=False,
    )`

It is also running for FFT (by using freqs instead of time) but I have not check the validity of the results yet.


BonneelP:

PR #367, opened 2021-04-06
H5py version >3 compatibility + save/load improvements

BonneelP:
Hello all,

This PR changes the way we save/load h5 files. h5py v3.0 includes some interesting improvements like "Reading & writing data now releases the GIL, so another Python thread can continue while HDF5 accesses data" (https://docs.h5py.org/en/stable/whatsnew/3.0.html) but it also changes how strings are handled. The requirements have been updated accordingly. I have run test_multi_multi save/load and compare the results (return empty list). "not long" tests are Ok as well. I haven't checked if we are still able to load a file generated with the previous version. 

I also included some miscellaneous modifications that improve the save/load of the corresponding objects (param_explorer.value not handle pyleecan objects, setter and getter have been changed to str for VarLoadCurrent....)

Best regards,
Pierre


BonneelP:
The code works with version 2.10 as well and I managed to save the results of test_multi_multi with 2.10 and load it with 3.2.1.

Issue #255, opened 2020-12-01
[CC] concentrated winding issue


SebGue:
Hello,

there seems to be an issue with the periodicity computation. I got a 20 pole / 18 slot machine that got a periodicity of 2, i.e. one periodicity is 10 poles / 9 slots.  Nevertheless `machine.comp_periodicity()` 'tells' that there is one periodicity with the a possible anti-perodicity.
Therefore e.g. the (automatic) winding is wrong (see fig. 1) and the simulation results are also invalid :-(

Fig. 1. Winding plot with wrong Phase A direction in slot 9 
![Winding](https://user-images.githubusercontent.com/43440731/100736786-8789c080-33d3-11eb-91fa-e10ea4d56ed3.png)

Best regards, Sebastian

BTW: It's great that the automatic winding definition is able to setup such pole and slot numbers already (despite the periodicity issue).

EmileDvs:
Hello @SebGue,

First you can check that the winding plot is consistent with the winding matrix. Then I think the machine.comp_periodicity() method  (which finds two anti-periods) is consistent with the winding plot, because the anti-periodicity is respected.

Finally, has the unit-MMF the proper polarity (p=9) ? It should not be the case, because you expect a different winding pattern to have so. 

For me it is more a problem of algorithm to define winding pattern given Zs and p (see #235).

Best regards,
Emile

SebGue:
Hello @EmileDvs,

you are right, it is not an issue with periodicity but an issue with winding pattern algorithm. I setup a user defined winding and every thing work as expected now.
So I will rename this issue respectively. Maybe someone of you could have a look at the algorithm to find the issue or at least add some comments.

I will leave this issue open and assign it to me to add some unit tests to avoid such issues.
E.g. check sum of conductors as in WindingUD and test for expected mmfs of common slot/pole combinations.

Best regards, Sebastian

SebGue:
Hello to you all,

@EmileDvs: Can you tell me something about the tooth winding algorithm. Is there a source or is it Eomys own work?

Actually I am not too happy with the Windings at all. There seems to be a lot of flaws.
So I got some ideas, I would like to discuss.
- It may make some sense to have a method 'convert_to_WindingUD' to easily setup a user winding for editing.
- There are at least 2 resources/tools to get winding patterns ([emetor.com](emetor.com), koil). So it would make some sense to have proper import for these tools.
- Maybe it would also make sense to create a collection with common ready to use windings that are accessable in the GUI.

[...]

Issue #235, opened 2020-11-13
[CO] Introduicing star of slot method for winding / New organisation for the windings


BonneelP:
Hello all,

We want to introduce the star of slot to compute the winding matrix but by doing so the current winding classes would be obsolete. So here is my proposition for the new organization:

- Winding is no longer abstract and is the default class to use.
- Winding would have two new properties Nrad and Ntan to tune the number of layer
- coil_pitch would be a property of Winding that can be used or not according to the number of layers.
- comp_connection_mat method would be added to Winding and return the star of slot matrix according to Nrad and Ntan (and qs)
- get_dim_wind method would be added to Winding and return (Nrad, Ntan)
- All the daughters of Winding would be removed except WindingUD and WindingSC (I propose to keep WindingSC because I expect it to evolve one day). 
- WindingUD would have some new properties like init_as_DW1L, init_as_CW2LR... that would set the user_wind_mat according to the result of the comp_connection_mat method from the removed classes (that way we can still use the old matrices).

Note that one day we can also introduce WindingUD.load_from_koil that would work as mentioned in #56 

What do you think of this new organization proposal ?
Best regards,
Pierre

EmileDvs:
Hello everyone,

We are digging up this topic as we would like to clean the winding user interface in the next release. 

We found a very powerful python package on gitlab (https://gitlab.com/martinbaun/swat-em/) that provides a python APÏ to generate any kind of winding using Star of Slots method (same as in Koil in fact). It also provides tools to import and export winding configurations, and algorithms to handle end-windings.

First I am going to implement the modifications that Pierre detailed above a while ago using the interface with swat-em to generate any kind of winding based on number of slots/poles/phases, coil span, number of turns per coil etc. 

Please let me know if you have any remarks,
Best regards,
Emile

SebGue:
Hello Emile,

I'm fine, thank you. Unfortunatly I couldn't contribute due to lack of time in the last weeks.

This is a very nice tool you found.
Is there a dedicated branch for integration of it? When do you plan to implement this package (to pyleecan/ to the GUI)?

Regarding end winding, it seems swat-em 'only' calculate the end winding connections?! So my end winding length calculation should be a complementation to it. 

Best regards, 
Sebastian

EmileDvs:
Hello Sebastian,

Thanks for the precision, I must say I didn't look much into the end-windings feature and it's good to know that it should be complementary to what you already did.

[...]

PR #378, opened 2021-04-28
Winding

EmileDvs:
This PR aims at solving issues #235  and #255.

The main changes in this PR are:

- all the changes listed in #235 original post by Pierre
- winding connection matrix can now be computed using Star of slots method based on SWAT_EM package
- merge of Stator Winding (1) and Stator Winding (2) widgets in machine GUI to ease winding definition
- renaming: Npcpp => Npcp and comp_Ntspc => comp_Ntsp to abide by notation standards

Still remains to do:

- update validation tests within GUI
- load in compatibility mode of old windings


BonneelP:
Hello all, 

Here is the current status of this PR: 
- Star of slot seems to be working in script and GUI (Winding pattern and Winding parameter are now one single step called "Winding")
- Technically all tests are Ok, but when checking the Tests/Results folder some images are strange
- All the existing machines are up to date and use the Winding object (Materials were also updated)
- Retro-compatibility works (a dedicated test was added)

So this PR is now Ok for review. Feedback is always welcomed to improve the modifications and to challenge the star of slot. 

The next steps are:
- Solving the images in Tests/Results
- Adding Import/Export CSV winding in scrip and GUI (we are currently discussing about the format to use, any input on this topic is welcomed)
- Maybe add the possibility to edit a user-defined winding in the GUI. 
- Check endwinding computation

Once the work will be finished we plan to let this PR open for some time before merging and then we should release pyleecan 1.3.0

Best regards,
Pierre

SebGue:
Hello Pierre,

updating the winding definition with such software as swat_em is a really great thing. 
However I got a remark. :-) 
I previously had some issue regrading unit testing and the advice of (I think) Benjamin was not to install PyQt5 and PySide2 at the same time. What do you think?
Further the reason not to have PyQt5 was because of licensing issues. Now what happens if some required package has PyQt5 as its requirement?

Best regards, Sebastian

SebGue:
Further ideas/ToDo's from my side:

- [ ] fix widget layout
- [ ] update figure of winding circuit with actual property names
- [ ] add fundmental winding factor to output
- [ ] it would be great if the coil pitch would be automatically set, i.e. pre filled

BonneelP:
Hello Sebastian,

Very important question that we are currently struggling with. The idea is to have matching version of PySide2 and PyQt5. We plan to update the requirements to have both installed with a fixed version. I currently have both PySide2 and PyQt5 installed and managed to run the GUI without problem. 

We had some issues with swat-em that uses PyQt5 even if we don't use the dedicated GUI. We have opened an issue to ask if we can contribute to swat-em to make it PySide2 compatible. The most significant issue we had was with matplotlib FigureCanvas. After the import of swat-em the FigureCanvas didn't act as a QWidget any-more and it introduces some bugs. For now we solve it by enforcing the import of the PySide2 objects at the beginning of run_GUI (cf commit c5fc531a7fa4e405097ccb74e68fd51cb8e3d06c). It is a workaround that we need to check works with pip install and pyinstaller (by waiting a correction with the contribution to swat-em). 

Maybe one day we can also investigate a parameter to switch between PySide2 and PyQt5 but I don't know if it can be done.

> fix widget layout
> update figure of winding circuit with actual property names

I guess this is the layout of the plot and the schematics. We are currently creating a new "vertical" schematics with the proper name and we plan to fix the figure/pixmap size once the image available. 

> add fundamental winding factor to output
> it would be great if the coil pitch would be automatically set, i.e. pre filled
> reintroduce radial vs. tangential layer selection

Interesting ideas, do you know how we could do that ? 

Best regards,
Pierre

SebGue:
For the winding factors there is a dedicated output in the swat-em GUI. So I guess there must be an api function for that as well.
With that we should also have a button to plot all the winding factors.
Regarding the coil pitch I haven't looked into the code where this could be done. Maybe there are some state variables that capture a change in number of slots and number of poles to trigger a calculation of the coil pitch.
The selection of radial vs. tangential layer should be a simple permutation of the winding matrix. But I haven't looked at the new winding code here as well.

BonneelP:
> The selection of radial vs. tangential layer should be a simple permutation of the winding matrix. But I haven't looked at the new winding code here as well.

Then we should do it as for Nslot_shift and is_reversed. The original matrix is preserved and changed when accessed. We can introduce a "is_reverse_layer" to reverse first rad layer and second rad layer and "is_change_layer" to convert tan layer to rad layer. SCIM_001 matrix is slightly different (it shouldn't be an issue), these options will enable to get exactly the same matrix as before
image

By the way, the strange images in Tests/Results folder are also in the current master so it is not linked to the PR. So all tests are Ok :) (We need to update test_FEMM_compare)

SebGue:
So CW will be tangential and DW will be radial by default, right?
Would you mind to introduce 'type_connection' 'not connected (nc)' e.g. as -1. I recently had such a machine.

BonneelP:
swat-em automatically select the most relevant between radial and tangential layers. The old algorithm are available as WindingUD methods. 

I'm adding the two options "is_reverse_layer" and "is_change_layer" in the GUI to be able to twist the matrix if the definition is not as expected. No problem to add the -1 to type_connection. I'm adding it to the csv.

For winding import/export csv we are thinking about: 
- One matrix per phase
- Each matrix separated by a blank line
- 1st column: Slot id
- 2nd column: Number of turn (+ or -) 0 for no winding
- 3rd column: Radial layer
- 4th column: Tangential layer

image
It will enable to update the winding matrix to have a different number of turn in some slots. 

Best regards,
Pierre

SebGue:
As for the export/import you could checkout the 'save_array'/'load_array' functions.
If you like you could also merge my 'EndWinding' branch for end winding length calculation. So this new release will be somehow 'complete' regrading windings. :-)

BonneelP:
> As for the export/import you could checkout the 'save_array'/'load_array' functions.

Yes I thought about these functions at first but I want the file to be as comprehensible as possible to enable editing. What do you think about only having the lines that actually contains winding to increase readability ? 
image

We can also add a first line with column labels.

> If you like you could also merge my 'EndWinding' branch for end winding length calculation. So this new release will be somehow 'complete' regrading windings. :-)

Yes if we can also solve the endwinding issue that would be great :) (I think that swat-em also handle end-winding but we need to be able to compute them for WindingUD as well).

SebGue:
Ah okay, now I got it. You compress the matrix by using the indices rather than the full matrix. I'm totally okay with that.
Maybe you could add a first comment line that has your explaination on the columns in it.

As far as I know, swat-em doesn't calculate the end winding length but will give the schematic, i.e. the end winding connections.

BonneelP:
After other discussions we think about this format:
image

with two options: 
- one to remove first column and first line
- one to remove the lines with Ntcoil=0

BonneelP:
Import / Export works in the GUI :) I don't think that the possibility to edit the matrix directly in the GUI is really useful since one can export the matrix, edit it in csv and import it back (maybe for later). As the tests are Ok the last open question is about End-winding. @SebGue Do you know where we can start ? Do you think that you can create a pull request with your branch ?

Best regards,
Pierre

EmileDvs:
@BonneelP, @SebGue,

I think we can merge this PR as it is, we already made a huge improvement concerning winding handling. 

Then I suggest to continue the work on calculating winding factors in a different PR, maybe @SebGue you can open a new issue about this ?

Best regards,
Emile

SebGue:
Hello Emile,

I would expect a plot (or a table) of windings factors as a function of wavenumber or electrical order (including sub harmonics). So we could keep this data as DataFreq or we call SWAT-EM to calculate it from the winding matrix calculation. (Although you mentioned, that we don't have access to SWAT-EM methods after discarding the respective object, there must be a workaround.)

For the coil pitch, I thought it would be a nice to have if there is a automatic prefill for a sensefull value in the GUI. Nevertheless I think there should be a getter for the coil pitch, since it is not an input property of all Winding classes.

Best regards, Sebastian

BonneelP:
Hello all,

I don't think I will have time in a near future to work on the winding factor plot. Is there anything else for this PR that we should investigate ? Otherwise I plan to merge it, close the old winding issue and open an issue on winding factor plot.

Best regards,
Pierre

PR #388, opened 2021-05-21
[NF] LSRPM machine

BonneelP:


Sijie-NI:
Hello, I am working for a new IPMSM with double windings (primary and damper winding). The objective of the damper winding is to reduce the flux density harmonics in the air-gap, thus the magnetic noise can be mitigated. The high frequency induced currents can be absorbed by the damper winding but the fundamental component will be prevented.

PR #397, opened 2021-06-02
[CO] Matlib reorganization

BonneelP:
Hello all,

I had some issues with the material library when checking the machine/library behaviour, so I took the opportunity to reorganize the code. The main modifications are:
- I have added a "ignore_list" parameter to compare so it is now way easier to compare the materials to find the differences
- Machine.get_materials_list is now get_materials_dict(is_unique) and returns the path to the materials which simplify a lot the get/set of machine materials
- I have replaced the Matlib object by its content: dict_mat (now material_dict) and all the material manipulation are handled in a single place: DMatlib.
- I have rename/remove several functions and property to be more consistent and understandable
- I have reorganized and added new tests to check that the behaviour is as expected: Tests\GUI\DMatLib\test_Workflow_Matlib.py now check edit, rename, delete, new, matlib=> machine and machine => matlib. 

I'm only talking about these modifications now because I didn't realized how much work I was about to do. The structure and usage of the code is mostly the same as before, the new functions enabled to do the same things with fewer code lines. The code should be easier to maintain. @SebGue I hope that these modifications are matching your usages of pyleecan, I'm willing to update them if necessary. 

Best regards,
Pierre


SebGue:
Hello Pierre,

I haven't look at your code but I got no special needs. The only nice to have feature is the capability to use more than one machine with the MatLib.
OT: At the moment I don't have too much time for contribution, so I hope you don't expect/need my input on any issue.

BR Sebastian

BonneelP:
Hello,

Thank you for the feedback. As I used the same principal: a dictionary with a key for the library and a key for the machine we can extend it to other machine. I just wanted to be sure that you had no particular constraints since I know that we talked several times about the Material Library. So if you have no specific needs I will merge this PR (anyway the test coverage is better than before).

Best regards,
Pierre

Issue #299, opened 2021-01-19
[NF] Magneto-striction


Superomeg4:
Hello everyone, 

I am going to start working on the implementation of magnetostrictive forces for one of my research project. 
The scientific background and the validation case is going to be based on the following article: https://ieeexplore.ieee.org/abstract/document/1284477

I am planning to create a dedicated class inherited from Force. Nevertheless, it is a little more complicated than the ForceMT module as it also depends on mechanical elastic properties. 

For the moment, I am planning to have an input dict for the required parameters, but maybe a dedicated object to store mechanical properties (such as Young modulus etc.) could be useful. 

Feedback welcome :-)

Raphaël

Superomeg4:
Progress report: 

The current work is available here: https://github.com/EOMYS-Public/pyleecan/tree/magnetostriction

A new class of Force (ForceTensor) has been created. It will compute nodal forces based on the input tensor model which will be a new class Tensor. In consequence, the output of this model is not AGSF but rather a MeshSolution object (which can be stored in OutForce).  At the end of this work, only one magnetostrictive tensor will be available.

PR #403, opened 2021-06-11
[NF] Magnetostriction nodal forces issue #299

jgdedamas:
Issue #299
Magnetostrictive forces have been implemented through a new simulation class : ForceTensor.
Comp_force_nodal, element_loop and comp_magnetostrictive_tensor are the main methods being used.
Some unit testing have also been added. 
A tutorial and for its use and a validation case is available in \Tutorials.


BonneelP:

Issue #407, opened 2021-07-09
Frame with Gap and structural Bars


ecs-kev:
Hi,

I commonly work with through flow ventilated machines, which generally have a gap between the stator and the frame. The gap is usually created by having rectangular bars between the stator and the frame. I noticed that it is possible to define a frame with a gap (though it doesn't seem to be possible using the GUI), but I could not find a way to define the geometry of the structural bars.

I would like to try to add that functionality. What would be the best way to implement this?
My plan was to add an additional class that inherits from Frame and has additional properties (Number of bars, Bar width) and methods (comp_height_gap).

Does this sound reasonable or is there a better/more preferred way to do it?

Cheers
Kevin

SebGue:
Hello Kevin,

your proposal is reasonable in my opinion. Depending on your actual needs, you may also try to use yoke notches (PR #376).

Do you need the frame and/or bars to be included in the FEA? This will require some further code extensions.
As for the GUI, I plan to release some extension to the GUI, that will allow editing generic pyleecan objects.

Hope this helps for now.
Best regrads, 
Sebastian

ecs-kev:
Hi Sebastian,

There's a few things I would like to do with his type of machine:
1) Calculate some geometric parameters like the surface area of the gap (this could probably be achieved using the notches).
2) Couple it with a CAD package to create 3D parts.
3) Run thermal FEA on it.

To achieve 2+3 i would be useful to have the bars implemented, though it might take a while until I actually get to do these steps.
I can't imagine the bars will have any impact on the electromagnetic FEA, so I wouldn't need them to be included for that.

Thanks for your feedback, I'll have a go at implementing the new frame type and see how I get along.
Looking forward to the new GUI extension. I haven't really gotten my head around how the GUI is implemented yet.

Kind regards
Kevin

BonneelP:
Hello Kevin,

To complete Sebastian's answer, the Frame class was intended to be a cylindrical Frame with Rint that matches the stator Rext. So a "frame gap" was not intended (hence why it is not available in the GUI) but it should be compatible. Creating a new class with bars would be great :) Do you have schematics or picture of this topology ? Maybe for later extensions we will have to make Frame abstract and rename the existing one to "FrameCirc" for instance. Anyway this class is not that much used so you should be able to add your modifications.

[...]

Issue #24, opened 2019-04-25
Method to split the Surface objects


BonneelP:
It would be great to be able to generically split the surface objects. It would have several applications:
- For now for the symmetry we split the lamination by assuming there is nothing on the "splitting line". In some cases there may be a ventilation, a notch or even a slot. Instead of changing the build_geometry it could be interesting to create the complete lamination surface and then "cut it twice" to get what is needed.
- For the build_geometry_wind we defined each surface manually. It could be interesting to just generate the complete winding surface then split it according to Nrad and Ntan. A second constraint would be to split it to have the same area for each surface. 

The method split could take as argument either two complex or a Segment object. In both case we assume that it define a complete straight line. The method would return two surfaces: the surface above the line and the surface below the line (can be None). The "above" would be defined by assuming the new coordinate system center is on the begin point of the Line and has the end point on the X+ axis.

BonneelP:
Next step is spliting an Arc when there are two intersections points (+ unittest for Arc2 and Arc3)
Then split method for surfaces 
We may need to introduce a "SurfRing": a surface defined by a external closed line (list of Line objects) and an internal closed line (to always have only one surface object for Lamination)

BonneelP:
This issue is still relevant. The build_geometry_active is using the split_line method of Surface but this method need to be improved: the argument "is_top" need to be removed to return always return both surfaces (to avoid calling twice the method).
Splitting lamination should already possible but a proper validation would be required.

BonneelP:
Hello all, 

I'm working on the next step for this issue in PR #435. Here is the status:
- split_line methods now returns both top and bottom part, which avoid double call and makes the code more readable. All related tests are Ok.
-Alpha0 was moved from VentilationCirc, Polar and Trap to Hole to be able to define an offset in any Hole (in particular for HoleUD). It is not meant to used in lam.hole
- The comp_point_ref of Lamination was corrected to take ventilations into account
- split_line method is now called to cut the ventilation (and not all the lamination surfaces as it would take too long and is not required most of the time) according to sym and draw_FEMM is correct (#266)
![image](https://user-images.githubusercontent.com/20533034/133815635-c22a5245-1855-4666-ad7d-a62974049c69.png)

[...]

PR #409, opened 2021-07-23
Frame gap

ecs-kev:


BonneelP:
Hello, 

Thank you for your contribution :) As a follow up to issue #407 here are some hints/remarks on the next steps for your PR:
- build_geometry, for further work, in some FEA software we may also need to return the air surfaces (in particular FEMM will need it).
- build_geometry_bar & comp_surface_bar, I don't know if it was intended but the bar looks polar instead of rectangular. Can you update the csv definition to specify if the bars are polar or rectangular ?
- build_geoemtry_bar, for now there is no symmetry in this function (but it can be tricky if the cutting line is on a bar). So we can leave it and wait for issue #24 
- comp_surface: to be consistent with the other method, can you update it to return the full surface (frame + bar) ? Then no more need for the comp_volume method.

Regarding the tests:
- For each method to test you need to define a function like in Tests/Methods/Machine/test_desc_dict.py. Pytest will understand that each function is a different test and if one fails it will run the others. To test your code, you can then add a "if __name__ == "__main__" and call directly the functions.
- The test for the three methods looks fine (for simple function a single call can be enough if there is no particular case). Can you add a test on comp_surface which is the generic methods that is used to compute the mass of the frame ? 
- To test the build_geometry functions, you can create a machine plot test like Tests/Plot/test_LSPM.py

Hope these few hints will help, feel free to ask otherwise :)
Best regards,
Pierre

ecs-kev:
Hey,

Thanks for your comments. I implemented them and will update the PR shortly.

Regarding the questions/comments:
1) I left this for now, but it should be straight forward to implement if required.
2) Yes, they are intended to be polar. I updated the csv defintion. In reality, they are generally rectangular and fit into a slot in the frame and stator, but I decided against implementing this for simplicity (and because the impact on modelling results will be negligible). 
3) I wasn't sure how to implement the symmetry, so I left it out. Will revisit it when #24 is completed.
4/5/6) Implemented all of these changes.

Kind regards
Kevin

BonneelP:
Hello, 

Sorry I just realized that this PR is still open. With all the modifications from your latest commit for me it is ready to merge. Is there anything else you wanted to add ? (For the model related method, we can use another PR)

Best regards,
Pierre

ecs-kev:
Hi Pierre,
No it's ready to go from my point of view. If you're happy to merge it as it is, please go ahead.
Regarding the model related method: Are you referring to the comp_height_gap_frame() method?
Kind regards
Kevin

BonneelP:
Hello,

By model related method I was referring to what you mentioned in the issue (CAD coupling and thermal computation). Basically, if we consider that this Pull Request "only" adds the geometry methods for the frame then we can consider it "merge-able". But indeed we can't do much with it for now so we can wait to have a proper application before merging it. 

I was just worried that you were waiting for the PR to be merged. If you prefer to wait, that's good for me, I will wait your next commits and green light to review/merge ;)

Best regards,
Pierre

ecs-kev:
Hi,
Would be nice to get it included as it is now, as I won't be able to get any further work done on this for the next few weeks as I'll be on holiday.
Cheers
Kevin

BonneelP:

Issue #413, opened 2021-08-02
[NF] DQ extension to DQH


BonneelP:
Hello all, 

We plan to extend the existing DQ transform methods to the DQH frame. 

Best regards,
Pierre

Sijie-NI:
Hello,
We can add one dimension in the transformation matrix as follows. P =2/3 for keeping the amplitude, P=sqrt(2/3) for keeping the power. Inversely P'=1 for the amplitude, P'=sqrt(2/3) for the same power.

![image](https://user-images.githubusercontent.com/64203774/128138369-0c2eef6f-b448-4c89-9eef-169c1d2f1ad7.png)

Sijie-NI:
Hello, I share the voltage equations on the dq0 frame:

![image](https://user-images.githubusercontent.com/64203774/128309705-de130df9-1487-430e-827f-1c414b35c915.png)

Best regards
Sijie

EmileDvs:
issue solved with #464

Issue #379, opened 2021-04-29
Electrical Equivalent Circuit for PMSM with Stator Damper Winding


Sijie-NI:
This project concerns a new electrical equivalent circuit (EEC) of a silent permanent magnet synchronous machine with a damper winding in stator slots. This project is based on a PhD thesis of the laboratory LSEE in the University of Artois "Damper winding for noise and vibration reduction of Permanent Magnet Synchronous Machine"

The EEC is separated into 2 parts, on the d-axis and on the q-axis:

d-axis:
![image](https://user-images.githubusercontent.com/64203774/116521274-109a1a80-a8d4-11eb-8050-bdbab7f89189.png)

q_axis:
![image](https://user-images.githubusercontent.com/64203774/116521363-29a2cb80-a8d4-11eb-9670-92e64ba00305.png)

SebGue:
Hello Sijie-NI,

very interesting project. Could you give an explaination on the EEC or a link to the related thesis?

Best regards, Sebastian

Sijie-NI:
> Hello Sijie-NI,
> 
> very interesting project. Could you give an explanation on the EEC or a link to the related thesis?
> Why is there a capacitance in series?
> 
> Best regards, Sebastian

Hello Sebastian, 

Thanks for your interest. 

I am PhD student at the University of Artois, who is working on this PhD thesis. The explanation will be completed later. I am learning how to model this machine in Pyleecan. We add this capacitor to form an RLC circuit in the damper winding, whose resonance frequency decides which harmonic could be reduced by damper winding. I have started my thesis in the laboratory LSEE. But I recommend papers and the thesis of my colleague Grégory Bauw, who applied this passive method in an Induction Machine, "Design of an induction machine with damper windings for noise and vibrations reduction". 

LSEE website:
http://lsee.univ-artois.fr/

Thank you very much

Sijie NI

Issue #372, opened 2021-04-26
Extension of Electrical Equivalent Circuits (EEC) to time harmonics


jlebesnerais:
The EEC implemented in Pyleecan (voltage to current transfer function) deal with fundamental values of voltage / currents, but when supplied with PWM or when introducing back emf harmonics the equivalent circuit does not work.
Extending EEC to time harmonics requires first to introduce skin effect and inductance factors for R, L elements depending on shape of conductor (round, square), without proximity effects for the moment.
Skin effects would also allow to improve winding loss calculations.

Sijie-NI:
Hello, 

I want to calculate magnetic fluxes Phid, Phiq (over time) from three-phase frame (PhiA, PhiB, PhiC), with Park transformation. Then, I load these fluxes in SciDataTool variable form (DataTime). Could I output directly the phase, module, frequency of their harmonics by "get_harmonics"? 

Best regards
Sijie 

def comp_EMF_harmonics(Phi_A, Phi_B, Phi_C,f, fs):

    Phi_d=sqrt(2/3)*(Phi_A*cos(delta)+Phi_B*cos(delta-2*pi/3)+Phi_C*cos(delta+2*pi/3))
    Phi_q=sqrt(2/3)*(-Phi_A*sin(delta)-Phi_B*sin(delta-2*pi/3)-Phi_C*sin(delta+2*pi/3))
    Phi_h=sqrt(2/3)*1/2*(Phi_A+Phi_B+Phi_C)

    time_axis = DataLinspace(
        name="time",
        unit="s",
        initial=0,
        final=(len(Phi_d) - 1) / fs,
        number=len(Phi_d),
        include_endpoint=True,
    )
    Phi_d_data = DataTime(
        name="Phi_d",
        symbol="Phi_d",
        unit="Wb",
        normalizations=None,
        axes=[time_axis],
        values=Phi_d,
    )

    Phi_q_data = DataTime(
        name="Phi_q",
        symbol="Phi_q",
        unit="Wb",
        normalizations=None,
        axes=[time_axis],
        values=Phi_q,
    )
    Phi_h_data = DataTime(
        name="Phi_h",
        symbol="Phi_h",
        unit="Wb",
        normalizations=None,
        axes=[time_axis],
        values=Phi_h,
    )

    # Number of harmonics
    N_harm=2000

    Phi_d_fft=Phi_d_data.get_harmonics(N_harm)
[...]

PR #416, opened 2021-08-06
[NF] GUI exe and installer

BonneelP:
Hello all, 

This PR introduces a new script to generate an executable to start the GUI and a dedicated installer. The idea is to add a new section on the website to upload an installer after each release. 

Best regards,
Pierre


SebGue:
Hello @BonneelP ,

another question: Are there any (recent) plans to extend the GUI beyond machine editing? Or what's the reason for releasing the exec. generator at all?

Best regards, Sebastian

BonneelP:
Hello Sebastian,

We are looking to ease the access to pyleecan. With the installer and the new button on the last step (cf #415) now pyleecan can be used as a mesh generator for electrical machine without having to install python or git. 

I take the opportunity to share our upcoming work on pyleecan:
- we are still working on the equivalent circuit for asynchronous machine (cf dedicated issues) that will reshape the Electrical module 
- we are planning to add the skew in MagFEMM and in the GUI
- we will add a GUI to SciDataTool (that will be available in pyleecan) to interact with the plot without knowing the syntax.

So no simulation GUI for now but if one day someone is interested to work of this topic, the executable will already be available.
Best regards,
Pierre

SebGue:
Hello Pierre,

so there are some things coming together... :-)
I recently thought about releasing a generalized gui (also for simulations) due to the features of #420. (I have also added a simple MeshSolution viewer internally.)

> We will add a GUI to SciDataTool (that will be available in pyleecan) to interact with the plot without knowing the syntax.

That's great, so I don't have to care about this, since this would be some of my next things to do. One thing to discuss here may still be left, but I will open a new discussion right after this comment.

> We are still working on the equivalent circuit for asynchronous machine (cf dedicated issues) that will reshape the Electrical module.

I thought I already introduced some EEC for IM, but I forgot :-) Do you mean #379 or is there another issue?

Best regards, Sebastian

SebGue:
... haven't tested yet, but is there also the ablility to export gmsh geo files, since meshes with general size assumptions may sometimes miss the intended purpose.

BonneelP:
Hello,

Yes most of the pieces are slowly taking shape to provide one day a simulation GUI :) We can also extend #415 to parametrize the exports.

Yes the first EEC was a great help but we need to go further. In particular #372, #413, #418
We still have a lot to experiment so we haven't fully decided what the architecture should be. Other issues should follow.

Regarding gmsh geo, yes that would be interesting. We have added the gmsh buttons to introduce this feature to the users. We know that we can go much further but we lack contributors on the topic :/ 

Best regards,
Pierre

Issue #419, opened 2021-08-30
Microsoft specific fonts used as default, and default_config_dict not changing font.


mjfwest:
It seems that Arial has been chosen as a default font, instead of a generic sans-serrif font.  Arial is not free and is therfore not installed on every platform. This generates a warning `[12:33:45] WARNING: Arialfont not available. Try: matplotlib.font_manager._rebuild()`.   (rebuilding won't work if you don't have the font)
Changing the font in `default_config_dict.py` does not fix the issue. 
`default_config_dict["PLOT"]["FONT_NAME"] = "FreeSans"`
further inspection of the codebase shows that `DEFAULT_FONT = "Arial"` is hard coded in `init_environment.py`, and also hard coded in many other plot functions and methods. 

Proposed fixes/questions:

- Outside the GUI, Is it really necessay for pyleecan to have an opinion on what font is used? 
- If it does, can it specify a generic font family such as 'sans-serif' and allow the platform to choose the installed equivalent font.
- If configuring a specific font is necessary for some reason, can the default_config_dict be used consistently?
-

mjfwest:
Would the following minimal change be accepted?  It would at least check to see what the default sans-serif font actually is, rather than assuming Arial is a default font.
```
from ..loggers import GUI_LOG_NAME
from ..default_config_dict import default_config_dict

# DEFAULT_FONT = "Arial"


DEFAULT_FONT = font_manager.findfont(font_manager.FontProperties(family=['sans-serif']))
DEFAULT_FONT = DEFAULT_FONT.split('/')[-1].split('.')[0]
DEFAULT_COLOR_MAP = "RdBu_r"

```

BonneelP:
Hello Martin,

Sorry for the long time to answer, I was in holiday. 
The point of config_dict["PLOT"]["FONT_NAME"] is to be able to select the font to use in all the graph/plot commands so it used in scripting as well. It is used to normalize the graph font for reports or publications. 

Normally to change the font you need to update the value in the main_config_dict.json file which is saved in the %APPDATA%/pyleecan folder. Changing the value in default_config_dict affect the original value when creating the main_config_dict when running pyleecan for the first time. 

[...]

PR #420, opened 2021-08-30
[NF] Generic Object Editor

SebGue:
This PR replaces the GUI's TreeView with a generic (pyleecan) object editor. The editor is related to #58 and mostly working.
Still there are some todos I got to implement, e.g. append/remove list or dict items, there is no link of the save state between both machine editors, ...


BonneelP:
Hello Sebastian, 

Sadly I won't have the time today to review your PR. I'm planning to do it next week to include it in release 1.4.0 (As #428 and #419 may also need further modifications). Next there will be the exe generation and the website update.

Best regards,
Pierre

SebGue:
Hello Pierre,

don't worry, since there are still some open points to this PR (see TODO's above).
Nevertheless there is one thing you could think about. This new 'tool' is pulling some informations about the classes (inheritance, properties, prop. types, ...) from the classes dict (i.e. Class_Dict.json) to be able to set the correct prop. type. E.g. if one would like to replace a stator slot, the editor will give the option to set SlotW10, SlotW11, etc.
Now while SciDataTool (and Mosquito) are compatible, I haven't had a good idea how to pull the same information from these packages to include them in the editor the same way. Maybe there should also be a ClassesInfo class with these packages?
If you understand this 'issue' and have some idea, let me know.

Best regards, Sebastian

BonneelP:
Hello Sebastian,

This PR is very interesting :) It will be really helpful to set the machine parameters that are not yet covered by the GUI. 
In particular, being able to interact with list/dict would enable to set the notches (on bore and yoke) which would be great ! What about having a line for each element of the list (hole then below hole[0], hole[1].... ?). Clicking on "hole" would enable to interact with the list length and then each line with the corresponding object.  

Regarding your issue with SDT / Mosqito, as we are going to add a GUI in SciDataTool it will more and more rely on the class generator. So in the near future we plan to extend it to add a "software" level to ease the generation process in all the packages (to remove progressively to links to pyleecan). We may even create a new separate repository for the class generator. With that in mind, we can assume that there will be a class_dict for each package which should be helpful ;)

Do you plan to continue the work on this PR or should we merge it and continue on another PR ?

Best regards,
Pierre

SebGue:
Hello Pierre,

you can merge this PR and I will increamentally implement the missing features.

The list/dict 'line per element' is already there. Only the add/remove is missing. 
I actually use this editor for all objects not only machines. You could 'hack' it and replace the object with maybe a XOutput object. But I'm not sure if you could do this after the GUI has started without crashing the app. So you may do this already at init.

A seperate generator sounds reasonable and the class_dict is really appriciated 👍 

BG S

SebGue:
Done for now. With the recent commit, it is possible to load any object to the TreeEdit. The MeshSolution viewer is also included.

BonneelP:
Thank very much for this contribution :)

PR #429, opened 2021-09-08
[NF] import partial winding csv

SebGue:
This PR enables to import winding definitons that are an integer fraction of the actual number of slots. So it is easier for users to define windings, since they only have to design the least common multiple of the winding.


BonneelP:
Hello Sebastian,

Thank you for this new feature, it will be helpful when importing winding for machine with too many slots :)
Do you think that you can provide a test for this new feature ? 

Best regards,
Pierre

BonneelP:
Hello,
To include it in the release along the SRM modifications I took the time to create the test. It will be included in the SRM PR.
Thank you for your contribution, with the "Empty" option a very small file can load a large winding :)

Best regards,
Pierre

PR #435, opened 2021-09-17
[NF] Ventilation and symmetry

BonneelP:
This PR is still work in progress, cf issue #24


BonneelP:
All the tests are Ok and I managed to run the following machine in FEMM:
image

SebGue:
So 'slit surface' is functional now?! That's huge.
Can bore shapes also been split now?

BonneelP:
In fact it was always functional (we already use it for winding layers), it was "just" waiting to be refactored and further used in the code ;)
At the end I don't "generate all then cut", I generate what is required and cut what is needed. So there are still some work to do on bore shape as explained in #24 but it is not that complicated to add. I will continue to work on this feature in the near future.

PR #444, opened 2021-09-28
[CO] loss test winding losses

SebGue:
This CO will improve the winding loss calculation in case there are asymmetric currents. (This isn't needed for the test itself, but will help if someone adapt the code from this test.)


BonneelP:
Thank your for your contribution :)

PR #450, opened 2021-10-04
[CO] Save json

BonneelP:
Hello,

This PR continues the discussion around the save function for json. 
- It should now be possible again to save object with no name property
- SlotUD/HoleUD now have a name property and are split in save
- indentation is now 2

@SebGue do you see anything else to fix/improve ? (I won't have time to work on "M400_50A_000001" to "M400-50A_Rotor-0")
Best regards,
Pierre

### DIFF
--- a/Tests/Functions/test_save_load.py
+++ b/Tests/Functions/test_save_load.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from os import getcwd, remove, listdir
-from os.path import isfile, join
+from os.path import isfile, join, isdir
 from unittest.mock import patch  # for unittest of input
 
 import pytest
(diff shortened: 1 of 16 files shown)


SebGue:
Hello @BonneelP 

have you discussed the default 'is_delete_old=True' with your collegues? I'm not sure about it, since it has some obvious risks.
Maybe old files could be moved to some subfolder by default and only get deleted if explicitly requested.

Best regards, Sebastian

BonneelP:
Hello,

For us when saving in a folder the folder should contains only the corresponding object that's why we set this behaviour. But I agree if someone save something else in this folder it can be harmful to delete directly the files. So I have introduced "type_handle_old" to choose the behaviour, the default is backup and we will use delete internally

Best regards,
Pierre

PR #460, opened 2021-10-15
[NF] Symetry for Bore notches in FEMM

BonneelP:
Hello all, 

This PR is a follow-up of the issue #24. I have updated the build_geometry of lamination to handle the symmetry on the notches by using the split_line method and here is the result (there are small notches on the rotor too):
image

The next step was to be able to have such topology in FEMM but I had again some boundary condition issues: 
- The airgap/sliding band side lines starts at Rbo (and not bellow)
- The shaft and Frame external radius have a particular BC and I don't know if it is compatible with something else than an Arc. 

So for now the yoke notches are not available for symmetry (by waiting further investigation) and for bore notches I have added dedicated surfaces to "close the void in the bore":
image
image
An advantage of using this method is that I was able to define bigger elements in the notches than in the Airgap. In the future, I think that we will add such surfaces to "close the slot" again to reduce the number of elements. 

To carry-on the work on issue #24 the next question are yoke notches and uneven Bore for LamHole.

Best regards,
Pierre


SebGue:
That looks great!
What is the actual issue with symmetry and notches in FEMM?

PR #470, opened 2021-11-04
[BC] Correct unit and name of Tem_slice and Phi_wind_slice in store

EmileDvs:
minor bug corrections


BonneelP:

PR #481, opened 2021-12-02
Electrical/EEC/Input rework (part 2)

EmileDvs:
This PR aims at improving the changes made on Electrical/EEC/Input classes in a recent PR. 

The main improvements are:

- LUT is now an attribute of EEC class (no longer of Electrical) and EEC.comp_parameters() look first to get parameter from LUT if it exists otherwise the parameter is calculated
- PWM workflow has been hugely improved and optimized, a dedicated validation test (`test_EEC_PWM`)  shows PWM voltage harmonics amplitude variation function of modulation index for sine and space vector strategies
- the EEC_SCIM using LUTslip has also a new validation case (`test_EEC_ELUT_SCIM_001`)
- OutElec has a new generic getter get_electrical() to get voltage and current objects
- DQH transformation now exists in frequency domain for voltage/currents defined as DataFreq and not DataTime (`test_dqh_transformation_freq`)


EmileDvs:
This PR is not ready to merge, tests still need to be corrected/updated

BonneelP:
All the tests are correct, I'm merging it to validate this new important step. 
Next step: LUT as a daughter of XOutput and we should be done with all this major rework (and it will be time for pyleecan 1.4.0 :)

SebGue:
I guess the next step is my part. Luckily I think I got some time to do that within the next maybe 2 weeks.

PR #485, opened 2021-12-14
Dxf gui optimization

EmileDvs:
This PR is not finished yet

This PR aims at optimizing Hole DXF import by selecting adjacent lines (in particular closed contours) with one click. 

It also optimizes other aspects such as searching the closest line to current mouse position and resorting liens after selection to create an ordered surface.

Also improve FEMM drawing after machine definition in GUI


BonneelP:
Thank you for this contribution, everything is way faster now :)

PR #489, opened 2022-01-20
[BC] comp_periodicity in WMachineTable

SebGue:


BonneelP:
Thank you for spotting this :)

PR #490, opened 2022-01-20
[BC] get stator current in WMachineTable

SebGue:
I haven't checked if the time is correct or if it even matter here.


BonneelP:
As it's only one time step it should be Ok, thank you for the correction :)

PR #503, opened 2022-03-01
Notch corrections for Lamination object

EmileDvs:
Few corrections to implement notches in Lamination object, especially LamHole rotors.

Also solve bug when importing FEMM files with worker parallelization.


EmileDvs:
The work is done, tests still have to be launched before merging.

Issue #498, opened 2022-02-24
[CO] EEC reorganization and LUT as a daughter of XOutput


BonneelP:
Hello all, 

As a follow up of our discussions in https://github.com/Eomys/pyleecan/pull/481 and https://github.com/EOMYS-Public/pyleecan/pull/12, we would like to gather in a single issue a fully detailed proposal on how to organize the Electrical module, EEC and LUT objects based on @SebGue idea to have LUT as a daughter of XOutput.

The first point is regarding the EEC objects. For now, to store the parameters we use a dictionary which is flexible but not standard. EEC_SCIM, EEC_LSRPM and EEC_PMSM are linked to specific and unique equivalent circuit with a frozen list of parameters. We propose to replace the parameters dict by the correct set of properties (R1, R2, L1, L2…) with the proper type, description, and min/max in the corresponding EEC objects csv files. 

Then we would use EEC objects both in Simulation.elec and in OutputElec (instead of the param dict). In a Simulation object, if a EEC value is at None it means that this parameter needs to be computed else it is enforced. comp_parameter would copy the simulation.elec.eec object and compute/enforce the corresponding parameters to get a full EEC (at the proper temperature and OP). So simu.elec.eec is the “initial state” of the EEC, Out.elec.ecc is the complete state for the corresponding temperature / OP. When a parameter is enforced, there is no computation to adapt the temperature and OP on this parameter, we assume that the user has provided the correct one. 

If new EEC or variation of the existing ones needs to be added, other objects with the proper properties/methods can be added (including by using polymorphism).

Regarding the use of LUT in Electrical module it would be as follow: 
- the property would be moved from simu.elec.eec.LUT_enforced to simu.elec.LUT_enforced
- LUT will contains an EEC at a given temperature and OP (known and stored in the LUT) and some tables (Phid, Phiq = fct(Id, Iq) for PMSM, Lm=fct(Im)  for SCIM)
- LUT would have method named “get_EEC(T=X, OP=Y)” that would adapt the parameter of its EEC to match T=X, OP=Y (if a parameter is missing from the LUT EEC, it would be computed directly). This method would encapsulate a call to eec methods with the same purpose adapted to the eec actual properties. 
- Electrical.run would first check for Elut to adapt a EEC, if not provided eec.comp_parameters would be called. In both cases a output.elec.eec is generated and output.elec.eec.solve can be called. 
- These modifications should simplify the comp_parameter methods

[...]

PR #506, opened 2022-03-03
[CO] EEC and LUT reorganization

BonneelP:
Hello all,

This PR is the second step for issue #498. The work is still in progress (in particular on EEC_PMSM) but the main modifications are taking shape and test_EEC_ELUT_SCIM_001 is running.
The EEC_PMSM and EEC_SCIM now have the parameters as properties and the run electrical solve a copy of the eec. 
The LUT is used to update the EEC to the correct OP and T. LUT now contains a EEC object and the LUT object was moved to Electrical.
We take the opportunity work on the skin effect factor but we can go further on this part.

Best regards,
Pierre


SebGue:
Hello Pierre,

I have just reviewed the PR and try to give you some feedback tomorrow.
Meanwhile, do you have some reference on the skin effect calculation and the some additional details on the different calculation types.
BR S

SebGue:
... could you also move/add the needed methods to abstract EEC class. (I don't know much about programming concepts, but this should be what you meant by polymorphism?!)

BonneelP:
All the modifications are done and the only failing test is the MTPA one that we will correct in part 3 of #498
We should merge and start part 3 on Monday

PR #507, opened 2022-03-07
LUT as a daughter of XOutput

BonneelP:
WIP


BonneelP:
All tests are Ok except MTPA one

SebGue:
So you still need PostLUT to setup a LUT?! I would prefer a dedicated setter method or the 'normal' way to setup a simulation.
Still I'm not sure about the LUT attributes (e.g. Phi_dqh_mag) in case a simulation is rerun. I.e. if one e.g. change mesh fineness and do 'LUT.simu.run()'. Then the LUT attributes won't be updated.

OT: Have you checked the load performance with the changes in import.

BonneelP:
Hello Sebastian,

> So you still need PostLUT to setup a LUT?! I would prefer a dedicated setter method or the 'normal' way to setup and run a simulation.

We continue to investigate the possibility of the LUT as daughter of XOutput an I starting to think about the following behaviour:
- Create a LUT of the correct type
- Call a method "set_default_simulation(machine, OP_matrix....)"
- LUT.simu.run
- Rather than creating a LUT object, the PostLUT would clean the non necessary contains of the XOutput/LUT object

Regarding loading performance, it depends on the size of the data stored within the LUT. Anyway we need to carry on the work on h5 save/load when we will have some time.

Best regards,
Pierre

SebGue:
Hello Pierre,

yes I saw the PostLUT cleanup. But I think this is the only benefit of PostLUT.
I think I got your described behaviour implemented in my proposal so far except the set_default_...

Regarding load performance I did some profiling and if I'm correct, the import of classes take the most time (independent of actual size of data). So I will redo this test with your PR to compare.

BTW: Did you see my PR on EEC_2? If you don't like it just close it (without worries :-) ).

Best regards, Sebastian

BonneelP:
Hello,

> yes I saw the PostLUT cleanup. But I think this is the only benefit of PostLUT.
> I think I got your described behaviour implemented in my proposal so far except the set_default_...

Then I will update the code to match this behaviour. (I will maybe not add the set_default_simulation at the end, there would be too many input arguments). Is there any other modification needed to ease your implementation ?

> Regarding load performance I did some profiling and if I'm correct, the import of classes take the most time (independent of actual size of data). So I will redo this test with your PR to compare.

We plan to check save/load h5 at the end of the month, all the data/information that you can provide on this topic would be very appreciated to help us decide what we need to optimize (maybe we can open a dedicated issue ?). I don't understand how "the import of classes" can take time, do you have any idea ?

> BTW: Did you see my PR on EEC_2? If you don't like it just close it (without worries :-) ).

Sorry for the PR, as you mentioned "without unit test". I thought that it was a Work in progress ^^" It's merged now thank you for this update :) Regarding your other PR, it is still WIP by waiting the merge of this PR ?

Best regards,
Pierre

SebGue:
Hello Pierre,

the set_default_sim. method is of course not mandatory, but could be helpful. Also I would not include all parameters but only the most important ones (e.g. current and speed,). But we can also add such method later.

For my PR, I think you can cancel that for now since you did the most important changes to EEC/LUT for now by yourself.
Nevertheless I may use some of the code to introduce an EEC_PMSM_2 class with my ideas later.

The only thing you could add to LUT methods would be an optional 'machine' argument to get fluxlinkage, inductance, ... for some similar machines (e.g. different length or number of turns). I will take some of the PR's code to introduce the respective functionallity.

For load/save, I can provide the profiling results. But I got some issues using h5 that I may need to investigate first.

Best regards, Sebastian

BonneelP:
I have updated the PostLUT to clean the current output and added the set_default_simulation method. I need to run the tests again and MTPA function still need to be updated

BonneelP:
The tests are Ok and we will solve the MTPA in another PR

PR #510, opened 2022-03-10
Improvement / bug correction in material library

EmileDvs:
This PR is still work in progress

- Account for temperature in remanent flux density in MagFEMM
- New getters get_Brm and get_Hc
- Correct confusion between mat.elec.alpha (temperature effect on resistivity) and mat.HT.alpha (thermal expansion coefficient)


EmileDvs:
This PR should be merged after #507

BonneelP:
Hello all,

This PR is ready to be merged (all Tests are Ok except MTPA that will be updated later). Before merging, @SebGue can you take a look at this PR (we went further than expected) ?
In particular we made the following changes:
- We have updated the materials and machine regarding the mat.elec.alpha (temperature effect on resistivity) and mat.HT.alpha (thermal expansion coefficient)
- We realized that the Prius magnet definition was odd. In fact the Hc value was the one for 60°C so we have introduced "T_mag" in Magnetics and Hc is not stored any more and we use Brm20 and get_Hc
- We have added several new magnets Magnet_40SH.json, Magnet_48M.json, ...
- For EEC_PMSM, FluxLink_FEMM and IndMagFEMM were replaced by a Magnetic object (MagFEMM in particular). And pyleecan/Methods/Simulation/EEC/comp_fluxlinkage.py compute the flux linkage using this Magnetic model (which enable to set each parameter of the MagFEMM object without duplicating the properties in FluxLink_FEMM and IndMagFEMM)
- We now only set the "Bore" label on the bore arc (and not on the slot/notches as before). It enables to define a better mesh in the airgap without having too many elements around the slots. 
- Now the GUI only creates material B(H) curve with an "ImportMatrixVal" object. "ImportMatrixXls" is still possible for retro-compatibility or scripting. The corresponding widget is now a generic DTableData and the WImportExcel is now more flexible and used by DTableData.
- We have introduce an artificial separation for the Magnetics parameters:
image
image
image
Artificial since its the same object and all the parameters can be set. We introduced this combobox because it was confusing to have on the same page magnet related parameters and the B(H) curve.

What do you think about these modifications ?
Best regards,
Pierre

SebGue:
Hello @BonneelP ,

you are working a lot on pyleecan lately. :-)
I will have a look. Could you tell me some deadline. 

Best regards, Sebastian

BonneelP:
Hello, 

We are about to release the next major version of Manatee so we are playing a lot with the architecture (this is also why we don't have time to update pyleecan website, tutorials, release currently but hopefully it should be possible soon). 

We are already moving forward with these modifications so there is no particular deadline ;) We are also thinking about removing the "is_isotropic" parameter from the MatLib GUI (since it can be different for mechanics and heat transfer). Isotropic material would be still available by setting X=Y=Z. We want to improve the clarity of the material definition and make sure that these new standards make sense and that it won't block you before merging it into pyleecan.

All the tests are Ok (after the latest modifications).

Best regards,
Pierre

BonneelP:
Tests Ok

SebGue:
Hello @BonneelP ,

I think I'm ok with your PR. I haven't checked the EEC part but I guess it is related to #498.

What is the purpose of the B(H)-Curve checkbox?
Where did you get the material properties from? Maybe we should introduce a 'desc' property to material to store some additional information.

Best regards, Sebastian

SebGue:
... don't know if this is related to this PR. Do you mind to allow all materials as 'Closing Material' for slot openings.

BonneelP:
> ... don't know if this is related to this PR. Do you mind to allow all materials as 'Closing Material' for slot openings.

Ha ha, we had the same return while presenting the feature internally :D We are about to change this feature to "Wedges material" ;) The "wedge area" will be the opening surface, later we can add other parameters to select another surface (in slotW10 we could have the "H1" part as a wedge and "H0" part as opening surface). It will be another PR. 

Best regards,
Pierre

BonneelP:
> What is the purpose of the B(H)-Curve checkbox?

Now that we have the combobox "material type" indeed we don't need this checkbox any more

> Where did you get the material properties from? Maybe we should introduce a 'desc' property to material to store some additional information.

Maybe @EmileDvs could answer from where the parameters are coming from. desc property already exist in fact, we can add it to the gui indeed.

EmileDvs:
> 
> > Where did you get the material properties from? Maybe we should introduce a 'desc' property to material to store some additional information.
> 
> Maybe @EmileDvs could answer from where the parameters are coming from. desc property already exist in fact, we can add it to the gui indeed.
>

Yeah we didn't tracked the origin of material properties, some of them come from material database on internet, other from validation articles or books etc. Indeed the best would be to review all material properties one by one and clearly write the origin in the description.

SebGue:
@BonneelP 

> Ha ha, we had the same return while presenting the feature ...

What does it take to change the material right now? Is it only GUI limited? 
I guess I need this feature really soon.

@EmileDvs 

Is there a good database on internet or is it rather a collection of companys datasheets?

BonneelP:
> What does it take to change the material right now? Is it only GUI limited?

I'm replacing "Slot.type_close" by "Slot.wedge_mat" and there are few other changes that needs to be done in FEMM. I plan to open the corresponding PR today.

PR #511, opened 2022-03-10
New electrical model for power input and loss minimization

EmileDvs:
This PR is still work in progress

- New loss model LossFEMM to calculate core losses, magnet losses, proximity losses, joules losses (cf https://www.femm.info/wiki/SPMLoss) from out.mag.meshsolution object calculated by MagFEMM
- New electrical model ElecLUTdq working with InputPower to achieve reference power and minimize losses based on LUTdq calculation and LossFEMM models to get losses

Validation cases upcoming


EmileDvs:
This PR should be merged after #510

SebGue:
Hello @EmileDvs ,

hope you have seen that LossModelBertotti is basically the same as the FEMM core loss calculation.

Best regards, Sebastian

EmileDvs:
> Hello @EmileDvs ,
> 
> hope you have seen that LossModelBertotti is basically the same as the FEMM core loss calculation.
> 
> Best regards, Sebastian

That's true, thanks for the input. Let me clarify this PR: this is not a final architecture of loss model that would obviously compete with the existing model usable from FEMM. I preferred to recreate this simplified loss model simulation apart from the existing Loss and OutLoss objects for two reasons:

- first to quickly progress on the InputPower object that will find the best (Id/Iq) given a LUTdq to achieve requested power while minimizing losses
- then to rely on the exhaustive validation case made by David Meeker to be sure that this simple loss model is well implemented

I am sure that we'll have some ideas later on to cleanly implement all loss models in LossFEMM in a way that they would be callable by Loss, and find again the results from David Meeker's validation case. This will be much better.

Best regards,
Emile

EmileDvs:
only Elmer test are failing so it should be OK for merge: 
image

PR #512, opened 2022-03-10
FEMM : Improving draw_FEMM by taking lamination symetry into account + Adding Closed Slot

BaptisteMadery14:


BaptisteMadery14:
Tested et ready for review

PR #518, opened 2022-03-25
[CO] Replacing type_close by wedges

BonneelP:
Hello,

@SebGue this is a work in progress (I need to work on the tests and the GUI) but the main modifications are there.
Can you take a look at pyleecan/Functions/FEMM/create_FEMM_materials.py to see if the material definition follow your needs ? I used the lamination material template but it may not be the most appropriate.

Best regards,
Pierre


SebGue:
This is really great! Lamination material is all I need since I would like to have slot wedges with powder iron with (nonlinear) low mu_r values.

BonneelP:
Normally I'm done with the modifications, we can add the wedge for all the slot (including UD) in the GUI. 
image
image
Default wedge material is the lamination one but another one can be selected.
For now only the "full wedge" is available but slot per slot we can introduce an option to select another area ("H1" part for SlotW10). This is why wedge widget is duplicated on all slots.
All the tests are missing, I will add them on monday. There is also a bug with Prius_DXF, as M400-50A is loaded as "old", the material is not selected in the combobox and the GUI is not displayed.

SebGue:
Hello @BonneelP 

again thank you for this really fast feature implementation.

Still I got some other issue, i.e. if I go back and forth in the GUI, the wedge is turned on automatically at some point.
Could you check if this is also happening at your side?

Best regards, Sebastian

SebGue:
Sorry, seems indeed that this is solved with your latest commits.

PR #521, opened 2022-03-29
[CC] Notches in LamHole

SebGue:
Add missing input parameter to LamHole.build_geometry to enable Notches.


BonneelP:

Issue #525, opened 2022-04-13
Jupyter notebooks, plotting and pyside2


mjfwest:
The compatibility with notebooks seems to be very broken now, which is a pity. 
I have recently upgraded from a fork based around 5daa0729338643e9a412c697c258b7be781162f8 where it did work, better at least. 

To get the notebooks in the tutorial directory to work one first needs to change  `%matplotlib notebook` to `%matplotlib inline`
and change all the calls to `.plot()` to  `.plot(is_show_fig=False)`

Once that is done `tuto_Machine.ipynb` works.
However `tuto_Operating_point.ipynb` fails with the error:
```
ImportError: Can't use Winding method comp_connection_mat: 
    Importing PyQt5 disabled by IPython, which has
    already imported an Incompatible QT Binding: pyside2
```

BonneelP:
Hello,

Thank you for having investigated this issue.

Sadly I haven't had much time for pyleecan recently (although several important PR were merged) but hopefully I will be able to come back to the project on next week and I will try to move forward to a new release 1.4.0 with all the latest modifications, a working SDT GUI and corrected tutorial/installation.

This issue is linked to swat-em (the package to compute the winding with star of slot) that requires PyQt5 (although we don't use its GUI). As pyleecan uses PySide2 it creates a conflict and the package can't be imported (in the notebook, it works in scripting) then no winding can be defined. In some part of the code we managed to correct it "the dirty way" by having a "from PySide2.QWidgets import *" before any import to swat-em to make sure that the widgets are the PySide2 ones and not the PyQt5 ones. We haven't had the time to investigate how to correct this import in the notebook nor how we could better handle this conflict but all help on the topic is welcome.

Best regards,
Pierre

mjfwest:
My fairly brutal solution so far, is to fork my own swat-em and comment out the plotting as it is not used?
this has moved me past that point, and it is now 
```
ImportError: Can't use MagFEMM method comp_flux_airgap: DLL load failed while importing win32api: The specified module could not be found.
```

This has worked previously.

mjfwest:
In response to my `DLL load failed while importing win32api:` , if using Anaconda, or miniconda the simplest solution is to pip uninstall win32 and reinstall with conda.
`pip uninstall pywin32`
`conda install pywin32`

It feels like this isn't getting to the source of the problem, but it works. 

https://stackoverflow.com/questions/58612306/how-to-fix-importerror-dll-load-failed-while-importing-win32api

[...]

PR #514, opened 2022-03-16
[NF] - Bayesian optimisation v1.1

HugoBecuwe:
First public push of the Bayesian Optimization method using the open-source library [SMOOT](https://github.com/RobinGRAPIN/smoot).

PR #529, opened 2022-04-14
[NF] Bore + Notch

SebGue:
This PR is still WiP, but I think the major first part of the work is done so far.


BonneelP:
Hello, 

I finally can take a day for pyleecan :) So I took the opportunity to check your PR, thank you for this work :) Here are few remarks to go further:

Regarding retrocompatibility, we check that the current version is before the first version with the new format. Next version of pyleecan will be 1.4.0 so in pyleecan/Functions/Load/retrocompatibility.py, Yoke_Notch_VERSION should be "1.4.0". It implies that everytime a retrocompatibility is needed a new release must be done (which is a good habit). 

Regarding periodicity, can you introduce a comp_periodicity_spacial method for BoreFlower that would return N ? For BoreUD we will later introduce a "per_a" property to manually set the periodicity.

pyleecan/Methods/Machine/Lamination/comp_point_ref.py: Do you think you can introduce a method "has_notch_yoke" and "has_notch_bore" or "has_notch(is_bore=True)" ? Same for "get_notch". 

Regarding the Tests here is the report: 
image

I plan to merge this PR, #514 and #511 before the release 1.4.0 and today I plan to work on #525 to get the notebook back.

Best regards,
Pierre

PR #537, opened 2022-04-25
[BC] set wedge material

SebGue:
fix a bug where lamination matieral BH curve was set for wedge matierals BH curve


BonneelP:
Thank you for spotting this typo :/

PR #552, opened 2022-07-12
[NF] Make the magnetic MeshSolution from FEMM compatible with slicing

RaphaelPile:
Make the magnetic MeshSolution from FEMM compatible with slicing if is_sliding_band = True


RaphaelPile:
@astonysheen That's for you

BonneelP:

Issue #104, opened 2020-06-15
Pyleecan under Ubuntu Linux - trying to get FEMM running on Wine  


normansdog:
It would be nice to have Pyleecan running on Linux but FEMM would have to run using Wine which needs to be tried out.

magnetron:
Alternatively, you could look into xfemm, which is supposed to run natively on Linux. 
Does anyone have an idea of how feasible it is to link to xfemm instead of FEMM?

BonneelP:
Hello,

Having the pyleecan coupling with FEMM working as well on Linux would be great. As for now FEMM is the only magnetic module, pyleecan is kind of "windows only" which it should not. There is a dedicated page on FEMM about how to make it run on Linux: http://www.femm.info/wiki/LinuxSupport

Indeed XFEMM can be a good solution (we should be able to reuse most of the existing code), but it would require to find a way to call it from python (I think that there is no "pyxfemm"). The xfemm provide code for C++ and Matlab/octave. When working on MANATEE (matlab) we used xfemm and there will be also the need to "clean" the geometry (remove duplicate lines and points). Anyway we need to work on such a feature for issue #16 as well (that was for xfemm that we started the "is_simplified" work). 

Does anyone knows what kind of modifications is required to call FEMM using Wine ? 

Best regards,
Pierre

crobarcro:
I am the author of xfemm. The problem with FEMM on Linux is that it cannot use ActiveX on Linux and instead uses a very slow and less reliable file based communication method. Otherwise it works surprisingly well on Linux. I'd happily accept contributions to xfemm to add python support, but have no time to develop it myself.

BonneelP:
Thank you for your feedback. Do you have any document/information about how much work it would represent to provide a python support to xfemm? We are not using too many femm methods so maybe we can start the work by focusing on what we are needing. Although, I don't think that I will have the time to develop it myself, maybe someone from the community can be interested.

BonneelP:
It seems that there was an update in Pyfemm in July 2021 that can help on this topic:
![image](https://user-images.githubusercontent.com/20533034/155946629-310a7e78-0433-4172-bf9e-1856bd7a8af7.png)
It could be worth investigating if it solves our coupling problem.

harshasunder-1:
Hello,
[...]

PR #560, opened 2022-08-31
basic linux support

harshasunder-1:
Hello!

this is with regard to issue #104 
The latest pyfemm supports linux, so I merged that into pyleecan.
From my tests, it supports drawing in femm properly, as well as simulation, though it is slow.

Some of the things that I noticed didnt work are :-
1)  the get_meshsolution function didnt work, but I think the errors were only to do with the filepaths.
Can we have a global isWindows flag somewhere? Maybe in the _FemmHandler itself?

2) parallelization didnt work - I think thats because in Linux we need to create a file in the same directory as femm to transfer the command, so if you open multiple FEMMs its not clear which instance is opening or deleting the file. A quick though ugly hack might be to just have multiple copies of the FEMM folder. 

Thank you
Harsha


RaphaelPile:
Hello there ! 
Thank you very much for this contribution. 
It is not really code related, but could you provide some details on how you managed to "install" FEMM on your Linux OS ? 

I am working with Mac OS, and I want to check if your PR is working on my side as well. 

Best,

Raphaël

harshasunder-1:
Hello!
I just installed it through wine. If you look at openFemm function inside  _FemmHandler.py file (in the Classes subdirectory), there are some default paths for linux that the code searches for FEMM (including the wine dir).  The new function even takes a path to the exe as an argument. Maybe for MacOS you might need to change those paths, or feed the path into the function as an argument
Thanks,
Harsha

RaphaelPile:
Okay thanks. For unknown reason, I was not able to use Wine. I am currently able to run femm GUI, but I still have to look how to make it work with pyfemm. I will let you know when I can test your PR.

BonneelP:
Hello, 

I checked your modifications and it looks fine to me. All the tests are passing on Windows including with parallelization.
I would only have two readability improvements to suggest: 
- What about renaming windowOS to is_window_OS ? 
- Do you think that you could split the modified methods into X_Win and X_Linux with an encapsulating method calling the correct version according to windowOS ? In particular for openfemm, callfemm and callfemm_noeval
I should have time to do both myself if needed. 

@RaphaelPile When do you think you would have time to check if it works on your side ? I'm currently not using Linux so I would be a great help if you could validate these modifications :) Otherwise I will take time to install Linux but I don't know when.

By curiosity what are the methods "X_setcomment" ? Are they methods introduced by pyfemm 0.1.3 ? I think we haven't updated the _FEMMHandler for a long time and I somehow managed to forget where this file came from in the first place :/

Thanks again for your contribution,
Pierre

harshasunder-1:
Hello, 
Sorry for the delay. I have made the changes you suggested, it makes the code much more readable. 
the new X_setcomment functions allow you to "set the comment text in the problem definition". They were adding in the latest pyfemm update
Unfortunately its still very slow in linux. You can use it to draw the machine, but very tough to use it for doing simulations and impossible for optimizations I think.  
Thank you,
Harsha

RaphaelPile:
@BonneelP I don't know, I have already issues to configure VSCode for python on my Mac OS. I will try my best to do it before the end of October.

BonneelP:
Hello,

I didn't manage to try your modifications on Linux, but all the tests are still Ok on windows and I'm about to create a new release to solve issue #569 so I'm (finally) merging your modification. Sorry for the long time to merge your contribution, it will be really helpful for the community.

I will still try to find some time to use your code on Linux to see if we can solve the speed issue.

Best regards,
Pierre

PR #563, opened 2022-09-14
[NF] BoreSinePole +  HoleM52R

SebGue:
This PR introduces a "new" hole HoleM52R alongside with a new bore shape. 
The hole is similar to HoleM52 but it has a rectangular void. Thats why I just added the 'R' to the name. But we can also change that for pure numeric HoleM names.
grafik

The bore 'BoreSinePole' is an adapted sine field pole geometry from the text book "Müller, Germar, et al. Berechnung Elektrischer Maschinen. Hoboken, NJ, United States, Wiley, 2008." ("Calculation of electric machines"). 
grafik

Both hole and bore can be utilized to get rotor geometries as the following.
grafik

Further the PR (#562 is included) add the BoreFlower.N GUI input. Otherwise only N=8 is possible due to default parameters. 

Hope this PR is interesting to someone. Feedback is welcome.

Best regards, Sebastian


BonneelP:
Hello Sebastian,

Thank you for your contribution :) There is no need to follow the HoleM5X format since it doesn't really makes sense (it is based on an old convention of Manatee V1 in matlab!). In fact we think about adding a "verbose name" to all the slot/hole (as for the magnets one) but clearly naming all the slot is tricky. 

Regarding BoreFlower.N we realized this very morning that we forgot this parameter ^^" Thank you for spotting that. We were considering enforcing it to 2*p rather than providing an input in the GUI but now that you have added the corresponding widget we will use that solution. Thanks again.

By the way I finally managed to update the website and to upload a installer for the GUI 🥳 I will try to investigate the PR this Friday but I don't know if I will have time to merge everything. Did you have the opportunity to check #558 ? It provide a notebook and reorganized some methods of the losses models (I didn't have the time yet to check the modifications in details).

Best regards,
Pierre

SebGue:
Hello Pierre,

the maximum radius of the bore shape is Rbo. If you got any idea to make that more clear, let me know.
Maybe there should be another marker at the inner dashed circle, which is Rbo.

I also didn't have time to check #558, but I will try to the next days. But first I got to review Elmer struct.
In that context, do you mind if I add optional radii to some of the other HoleM's similar to HM52R. Otherwise structural simulation results will have really high stress values in the void corners (due to singularity).

Best regards, Sebastian

BonneelP:
All clear for HoleM52R :)
I still need to check BoreSinePole and add the corresponding tests.

BonneelP:
Hello Sebastian,

I finally managed to add the missing GUI test so we can merge this PR. At first I thought about editing the schematics but while reading the code, I don't think that we can do better that what you already did. I have added the reference in the code for clarity.

Thank you for your contribution and sorry for the long time to merge.
Best regards,
Pierre

PR #565, opened 2022-09-27
[CC] allow ForceTensor.group=None

SebGue:
... as described in ForceTensor csv (group: "Name of the group selected for magnetic force computation. If None, all the domain is selected.")

### DIFF
--- a/pyleecan/Methods/Simulation/ForceTensor/comp_force_nodal.py
+++ b/pyleecan/Methods/Simulation/ForceTensor/comp_force_nodal.py
@@ -28,8 +28,11 @@ def comp_force_nodal(self, output, axes_dict):
 
     meshsolution_mag = output.mag.meshsolution  # Comes from FEMM simulation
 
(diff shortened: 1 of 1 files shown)


RaphaelPile (approved):

SebGue:
This issue leads to the question if we should change MeshSolution.get_group behavior?
At least there should be some test if get_groups input arg 'group' is a list of str or a str and raise an error if it is not.

Best regards, Sebastian

RaphaelPile:
I guess you are right. For exemple in the pyvista plots we are already checking if there is something in the requested group list before calling get_group. I think this behavior should be generalized, it would be more robust.

RaphaelPile:
@BonneelP I am not able to launch the tests on my MAC (yet), but I think this PR should be fine ... maybe some one or two small issues with the plots.

BonneelP:
Hello,

Test and review Ok, thank you for the contribution :)

Best regards,
Pierre

PR #566, opened 2022-09-29
[CO] Modify load_json to make sure that we open the file with the right encoding

BaptisteMadery14:


BonneelP:
Already done in PR #558

Issue #569, opened 2022-11-12
Problem installing requirements.text


babdwahab:
Hi, 
I am new to the pyleecan world and am really excited to use and contribute. I have installed the GUI (method 01). But I want to use it in Jupyter Notebook as it is more flexible. Following the prerequisite on the website (https://pyleecan.org/prerequisite.html#installation) my cmd shell successfully installed all the requirements with some error:

ERROR: Ignored the following versions that require a different python version: 1.21.2 Requires-Python >=3.7,<3.11; 1.21.3 Requires-Python >=3.7,<3.11; 1.21.4 Requires-Python >=3.7,<3.11; 1.21.5 Requires-Python >=3.7,<3.11; 1.21.6 Requires-Python >=3.7,<3.11; 1.6.2 Requires-Python >=3.7,<3.10; 1.6.3 Requires-Python >=3.7,<3.10; 1.7.0 Requires-Python >=3.7,<3.10; 1.7.0rc1 Requires-Python >=3.7,<3.10; 1.7.0rc2 Requires-Python >=3.7,<3.10; 1.7.1 Requires-Python >=3.7,<3.10; 1.7.2 Requires-Python >=3.7,<3.11; 1.7.3 Requires-Python >=3.7,<3.11; 1.8.0 Requires-Python >=3.8,<3.11; 1.8.0rc1 Requires-Python >=3.8,<3.11; 1.8.0rc2 Requires-Python >=3.8,<3.11; 1.8.0rc3 Requires-Python >=3.8,<3.11; 1.8.0rc4 Requires-Python >=3.8,<3.11; 1.8.1 Requires-Python >=3.8,<3.11
ERROR: Could not find a version that satisfies the requirement PySide2 (from versions: none)
ERROR: No matching distribution found for PySide2

Then I saw a post on Git that the requirements are (https://github.com/Eomys/pyleecan/blob/master/requirements.txt). when I do this now I have a different problem:

C:\Users\Admin>pip install -r C:/Users/Admin/Desktop/Pyleecan/requirements.txt
Collecting git+https://gitlab.com/CedMrnl/swat-em.git (from -r C:/Users/Admin/Desktop/Pyleecan/requirements.txt (line 16))
  Cloning https://gitlab.com/CedMrnl/swat-em.git to c:\users\admin\appdata\local\temp\pip-req-build-mh_dgkaw
  ERROR: Error [WinError 2] The system cannot find the file specified while executing command git version
[...]

PR #570, opened 2022-11-16
[BC] Solve pip install issue #569

CedMrnl:
As swat-em default package [doesn't respect Apache 2 licence due to PyQt5 in requirements](https://gitlab.com/martinbaun/swat-em/-/merge_requests/12), we use a [temporary fork](https://gitlab.com/CedMrnl/swat-em.git). 
However, PyPI cannot depend on packages which are not also hosted on PyPI. 

To solve this issue, I propose to call `pip install git+https://gitlab.com/CedMrnl/swat-em.git` directly in setup.py

Best regards, 
Cédric

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,9 @@
+from os import system
+from sys import executable
+
 try:
     import setuptools
 except ImportError:  # Install setuptools if needed
-    from os import system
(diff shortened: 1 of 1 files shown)


BonneelP:

PR #589, opened 2023-01-06
[CC] allow dict keys with backslashes in h5

SebGue:
In HDF5 groups are similar to file paths, i.e. there are backslashes in group paths.
This will cause issues if a dict has keys with backslashes in it. E.g. in an optimization where the objective is some ratio "x/y", the objective is stored in xoutput_dict with key "x/y". Then the result is saved without errors but can't loaded properly due to corrupted groups.
This PR fix this issue by converting backslashes in keys to "\\\x2F" and back.

### DIFF
--- a/pyleecan/Functions/Load/load_hdf5.py
+++ b/pyleecan/Functions/Load/load_hdf5.py
@@ -74,6 +74,9 @@ def construct_dict_from_group(group):
             # Check if key is an int
             if is_int(key):
                 key = int(key)
+            # Convert key in case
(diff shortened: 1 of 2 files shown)


BonneelP:
Hello, 

Sorry for the delay, we had a lot of work on our commercial software recently but the release is now done so I (finally) have more time for pyleecan. 
Thank you for spotting this error :) I will add a test in the master branch on this case. 

Best regards,
Pierre

PR #600, opened 2023-02-15
Improvements on creation of geometry and meshing by GMSH

ajpina:
This pull request will improve the geometry and mesh generation through gmsh. The code has been rewritten to use OCC kernel instead, and this allows to perform boolean operations for the better creation of surfaces in the air gap. The sliding band is being drawn in such a way that is compatible with the magnetics solvers in Elmer fem.
Additionally, this pull request will allow to create and mesh almost all the example geometries available in the database: 
 pyleecan/Data/Machine .
There will be future improvements to support outer rotor topologies. Currently, the outer rotor geometries can be drawn without sliding band.
I am also trying to identify and fix some of the boundary condition settings that seems to be inconsistent with how those are currently treated in pyleecan.
GUI has a button to create gmsh file and default to sym=1, it would be good that defaults to is_sliding_band=False and is_airbox=False too.

### DIFF
--- a/pyleecan/Functions/GMSH/draw_GMSH.py
+++ b/pyleecan/Functions/GMSH/draw_GMSH.py
@@ -49,7 +49,7 @@ def draw_GMSH(
     is_lam_only_R=False,
     user_mesh_dict={},
     path_save="GMSH_model.msh",
-    is_sliding_band=True,
+    is_sliding_band=False,
     is_airbox=False,
     is_set_labels=False,
     is_run=False,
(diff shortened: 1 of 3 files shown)


BonneelP:
Thank you for your contribution, this looks very interesting :) Starting from next week I should have more time for pyleecan (we are about to release the next version of our commercial software). I will try to review your PR (and the other pending one) by then. @SebGue Do you think you can take a look as well ? 

Best regards,
Pierre

BonneelP:
Hello, 

I'm currently reviewing this PR and I find few things to correct but otherwise its a great step forward thank you :)
I miss clicked and merge the PR a bit too fast. I will open another PR with my modifications to carry on the discussions if needed. 

Best regards,
Pierre

PR #612, opened 2023-03-16
Winding definition improvements

Eomys:


BonneelP:
Hello all,

With this PR we will rework how the winding are defined in the GUI to be more understandable and flexible. Here are some of the modifications that we will investigate:
- N*3 phases
- delta / star connection
- Step "Stator Lamination" before "Stator Conductor" to have L1 set to be able to compute winding mass/resistance in Conductor
- Few renaming for more clarity
- Plot radial pattern and Plot linear pattern from swat-em
image

Best regards,
Pierre

Ps: I think I will wait to end this PR before releasing 1.5.0

PR #624, opened 2023-05-18
GMSH and Elmer

ajpina:
This is a follow up on PR #600 . The issue with boundary conditions on lines is fixed in this PR, and "test_magelmer.py" has been tested as well.
This PR uses OCC engine in Gmsh to perform boolean operations for drawing airgap and sliding band surfaces. After performing CUT operations, new resulting lines are searched for and set as physical lines to be used for MagElmer.
Most of geometries in the database are supported and can be converted into gmsh with any applied symmetry. However, Airbox region is still work in progress when geometry is drawn with sym=1.
Best,
Alejandro


BonneelP:
Hello,

Thank you for your PR, I will try to review it as soon as possible but most likely not this week.

Best regards,
Pierre

BonneelP:
Thank you for your contribution and sorry for the merge delay :)

PR #640, opened 2023-08-31
Slot M19

clementd08:
Added new Slot M19.
Corresponding to a Trapezoidal Magnet with polar top
SlotM19


BonneelP:

PR #648, opened 2023-09-18
[NF] wedge type to SlotW14 and SlotW29

clementd08:
Hello all,

This Pull Resquest introduce a new methods wedge_type. Now for the SlotW14 you can choises if the surface of wedge is full or not.
Like you can see on pictures. 
SlotW14_wedge_type_1
SlotW14_wedge_full
We have choise if the wedge is full, wedge_type = 0 and if it's like the first picture wedge_type = 1. 
But, wedge_type isn't a booleen to have the possibility to add other type.

The same for the SlotW29
SlotW29_wedge_type_1

Best regards,
Clément


BonneelP:

PR #651, opened 2023-09-22
[BC] Error with closing lines for slot with tangential winding

BonneelP:
Hello all,

We found a very specific bug with the FEMM coupling, 2 layer tangential winding and some particular slots with overlapping winding/slot lines: When a slot have a double layer tangential winding and that a line perfectly overlap the opening/active zone the copy/rotate method of fast draw fails (the closing line is missing):
260493120-8c17ca81-7627-43ff-a58a-938531b01ec6

The reason is that FEMM selects the surface to copy rotate by selecting all the lines of the surface by selecting the closest line to the center of the line. As the slot and winding line overlap there are in fact two lines to select and if the slot line is longer that the winding line, then only the slot line is copy/rotate and the winding surface is not closed ! Ex: SlotW22 with W0 < W2*0.75

Best regards,
Pierre


pyleecan/Methods/Slot/SlotW21/build_geometry_active.py
```diff
@@ -85,6 +88,19 @@ def build_geometry_active(self, Nrad, Ntan, alpha=0, delta=0):
             )
             surf_list.append(surface)
 
+    # Correct bottom line for particular case (cf Tests\Validation\Magnetics\test_FEMM_fast_draw.py)
+    if Ntan == 2 and Nrad == 1:
```
AdrienLeduqueEomys, line 92: Est ce que la condition sur Nrad est nécessaire ?
Que se passe il si Ntan >2

BonneelP: En gros 99% des cas c'est (Ntan=2, Nrad=1) ou (Ntan=1, Nrad=2) ou (Ntan=1, Nrad=1). Dans la GUI on ne peut pas faire autre chose. Sur certains slots simples l'algo de build_geometry_active a été pensé avec Ntan/Nrad quelconques pour laisser étudier des cas exotiques. Donc j'ai laissé de côté le correctif (complexe) sur ces cas purement théorique. Pour corriger dans ce cas là il faudrait trouver la bonne surface à corriger.

AdrienLeduqueEomys: Dac

AdrienLeduqueEomys:
- [ ] Review

PR #660, opened 2023-10-16
[NF] Renaming Wmag, Hmag, Hkey and Wkey for W1 and H1

clementd08:
Hello all,

This PR introduces a new naming to simplified test and check. 
Wkey/Hkey was just present in GUI so we have decided to modified that, to simplified code.

Best regards,
Clément


BonneelP:

